PR #14, opened 2017-10-23
Updated gitignore and added potion effect thing

qkleinfelter:
no clue if potion effect dealio works


kerbybit (changes requested):
getActivePotionEffects should return a String[] instead of an ArrayList<String> to play nicer with nashorn

You should also add it as a var to providedLibs and keep it updated every tick

PR #15, opened 2017-10-24
Added player armor getters

qkleinfelter:


mattco98:
Inside each function, you have to ensure that `mc.thePlayer` is not null. Otherwise, the game will instantly crash as soon as it starts (because `mc.thePlayer` is null while in the main menu). See the other functions in that file as examples.

qkleinfelter:
ok this should work now @FalseHonesty

qkleinfelter:
also in my last commit i changed the armor getters to return the display name instead of the toString

qkleinfelter:
@FalseHonesty @kerbybit

qkleinfelter:
Closing cause matt basically rewrote my stuff but betterer

PR #19, opened 2017-10-29
This has 15 commits but only one is real so do like matt says in the ss in the comments, essentially I made getPlayerYaw() slightly more efficient

qkleinfelter:
http://prntscr.com/h3bipt the screenshot that matty boi says is good @FalseHonesty idk how to assign u so i @'d u


FalseHonesty:
how have you managed this

PR #25, opened 2017-11-10
Fixed /ct simulate and /ct dump

mattco98:
Also reverted the thing kerby changed in `dumpChat()` cause ChatLib automatically adds colors.


FalseHonesty:
I sure hope this works

PR #76, opened 2018-02-19
Added documentation to some methods in ChatLib and RenderLib

qkleinfelter:
I know it looks like a massive pull but in actuality its like 27 lines of changes just a bunch of reverts happened in my prior attempts at helping


kerbybit:
You dont need to say that the separate downloadImage methods are "versions" of the original, just give them separate descriptions.

the comment on drawImage that you added should either be a double slash comment or specifically state that it is not meant to be used by the user (they don't have access to the ResourceLocation object) see [here](https://www.chattriggers.com/javadocs/com/chattriggers/ctjs/minecraft/wrappers/objects/Block.html#Block-net.minecraft.block.Block-) in the block construstor

getChatLines returns 1000 entries max, not 100

kerbybit (approved):

PR #122, opened 2018-05-25
Add getEntity to the Entity wrapper

WalshyDev:
This is so that I can access net.minecraft Entity specific methods.


FalseHonesty:
Just add a `@Getter` to the entity field :)

WalshyDev:
Yeah I was thinking that looking through other classes.

WalshyDev:
By the way I can't sign the CLA currently, seems to be some issues. It seemingly kept expiring the token on approval and trying to then make me re-authorize. Now GitHub has flagged it as suspicious abuse.
https://bots.are-pretty.sexy/20ba84.png

FalseHonesty:
Thank you for your contribution! <3

PR #130, opened 2018-08-04
Added new ChatLib.editChat methods

ThePoptartCrpr:
Added ChatLib.editChat(int chatLineId, String toReplace) and ChatLib.editChat(int chatLineId, Message toReplace)


FalseHonesty:
Closing because already implemented

PR #140, opened 2018-08-27
Cleaned up some docs

mattco98:


kerbybit (changes requested):
Anything that returns a Minecraft object (like Client.getMinecraft) should still specify what it is returning in the docs to make it clear what it is

kerbybit (approved):

PR #141, opened 2018-08-29
Started converting triggers

mattco98:


FalseHonesty:
See comment left on #143

PR #151, opened 2018-11-11
Promote https

RDIL:


mattco98:
This actually breaks tinypic. If you open the contributing in your repo, you can see that all of the https tinypic urls don't load. What you could do is upload all the images to imgur and then replace all of the links with those urls.

RDIL:
oof ok

FalseHonesty:
@RDIL yeah

kerbybit:
We should probably just upload these images directly to the ct website

RDIL:
I’ll put it on imgur later for ya guys

PR #154, opened 2018-12-01
Remove .INSTANCE from Tessellator in providedLibs

ThePoptartCrpr:
Should fix Tessellator issues.


ThePoptartCrpr:
(matt's committing instead apparently)

PR #171, opened 2019-09-17
added client side command running

Heknon:


src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/ChatLib.kt
```diff
@@ -85,6 +86,18 @@ object ChatLib {
...
     fun command(text: String) = say("/$text")
 
+    /**
+     * Runs a command.
+     *
+     * @param text the command to run, without the leading slash (Ex. "help")
+     * @param clientSide should the command be ran as a client side command
+     */
+    @JvmStatic
+    fun command(text: String, clientSide: Boolean) {
```
FalseHonesty: Kotlin has no need for overloaded functions. Rather, add the clientSide parameter with a default of false to the existing `command` function. Make sure to annotate that function with `@JvmOverloads` so it can be called without the optional parameters from JS.

FalseHonesty (changes requested):

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/ChatLib.kt
Heknon: Fixed and tested. All working.

FalseHonesty (approved):

FalseHonesty (approved):

PR #177, opened 2019-12-06
Adds option to specify image path

Heknon:


kerbybit:
you can already do that though

PR #193, opened 2020-07-27
Added function ChatLib.addFormatting

glasss13:
It literally just didn't exist even though it was referenced in other places in the code.

### DIFF
--- a/src/main/resources/js/moduleProvidedLibs.js
+++ b/src/main/resources/js/moduleProvidedLibs.js
@@ -95,8 +95,10 @@ global.register = function (triggerType, methodName) {
 
 // String prototypes
 String.prototype.addFormatting = function () {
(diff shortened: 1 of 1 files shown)


mattco98:
The function is far too naive (think `ChatLib.chat("this & that")`). Also, the function already exists: `ChatLib.addColor`. As far as I can tell, `addFormatting` is only referenced once. Let's just replace that instance to use the existing function :)

glasss13:
Ah yea, didn't think about that. Probably shouldn't be committing this late. Also yea didn't even notice the addColor function.

mattco98:
Well, you can't change the name of the user-exposed function in providedLibs, as that could be breaking. You could definitely add it as an alias though.

glasss13:
I'm facepalming so hard rn at myself.

PR #194, opened 2020-08-02
Fix playerLeft event not triggering on player leave.

glasss13:
World.getPlayerByName returns null if player isn't found.


FalseHonesty:
That's a sneaky one, I guess the `World.getPlayerByName` API used to throw instead of returning null huh? Good find :)

glasss13:
Yep! [https://github.com/ChatTriggers/ChatTriggers/commit/59c49add76ad36f219f9f105d213e717ab91e245#diff-7f4cad0816a68956e3a09a7ac0d405f4](https://github.com/ChatTriggers/ChatTriggers/commit/59c49add76ad36f219f9f105d213e717ab91e245#diff-7f4cad0816a68956e3a09a7ac0d405f4)

PR #196, opened 2020-08-19
Allow for module creators to add a "helpString" to metadata

qkleinfelter:
This string then gets printed to the user when they import the module.


mattco98:
There are two problems here:
1. To avoid spamming the console, we should probably put some limits on the contents of the string. I'm thinking three lines is a good limit, so limit it to `3 * <max chars per line>` characters. I can't remember how many characters there are per line in vanilla MC. 
2. Experienced players that don't need the assistance should be able to prevent this from being printed to the console via a config option. 

Also, I don't really like the name `helpString`, lets go with `helpMessage` unless you can come up with something better than that.

PR #197, opened 2020-08-28
Fixed issue with packetReceived trigger causing crashes.

glasss13:
Due to some weird reasons with how minecraft works with threads and packets the old packetReceived trigger would sometimes cause crashes. This was solved by changing the injection location.


Debuggingss:
Great fix! Thanks

PR #205, opened 2020-10-13
fix: stop no connection from freezing the client on launch

ghost:
decided to do this after yesterday's issue where ct would freeze the client if it couldn't connect to the servers


Sk1er:
I'm here to say hi

FacioZMafii:
k

PR #211, opened 2020-12-07
Fix packetSent Trigger Causing Failed Assertion Errors

glasss13:
If the packet in question is not in the correct thread a scheduled task is created to make it execute in the correct one. Trying to access the packet before this has happened seems to cause Failed Assertion Errors. This commit also changes packetReceived to be in the same place as packetSent. Unfortunately, the amount of injections is still two because some packets are sent already in the correct thread which requires another injection.


glasss13:
Very confident this fixes assertion error but I still want to do more testing because apparently it can at times takes up to 2 hours for the bug to manifest.

FalseHonesty:
Sooooooo, does this work?

glasss13:
Closing because #232

PR #217, opened 2021-01-08
New Javadocs using dokka

DJtheRedstoner:
Userdocs task will now generate dokka html docs. You can see an example hosted [here](https://djtheredstoner.github.io/chattriggers/javadocs/chattriggers/index.html). In order to get dokka to work I had to update to gradle 5. This meant switching to replaymod forgegradle. If anyone knows a better way to do the package exclusions please let me know since I spent ages trying to get it to work but i'm sure there is a better way.


DJtheRedstoner:
superseded by #270

PR #226, opened 2021-03-07
Build: Fix Dependency Errors

glasss13:
Spongepowered maven repository was moved and now requires https
connection.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,11 +12,11 @@ buildscript {
         mavenCentral()
         maven {
             name = "forge"
-            url = "http://files.minecraftforge.net/maven"
+            url = "https://files.minecraftforge.net/maven"
         }
         maven {
             name = "sponge"
-            url = "https://repo.spongepowered.org/maven"
+            url = "https://repo.spongepowered.org/repository/maven-public/"
         }
         maven {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -95,7 +95,7 @@ repositories {
     jcenter()
     maven {
         name = 'sponge'
-        url = 'http://repo.spongepowered.org/maven'
+        url = 'https://repo.spongepowered.org/maven'
```
DJtheRedstoner: This should be changed to `https://repo.spongepowered.org/repository/maven-public/`, the new repository, the old url is not supported at all.

DJtheRedstoner (changes requested):

build.gradle
glasss13: Oh, yea, reading probably would've been helpful. I assume that then also the reference on line 19 needs to be changed as well?

DJtheRedstoner: Yes, any and all references should be changed (and you should also look for other maven repositories using http and try to change them to https)

edit: just noticed the forge maven in buildscript block is http, this should be changed.

glasss13: Got it, thanks!

DJtheRedstoner (approved):
Looks good now!

PR #229, opened 2021-03-19
Add module update message

camnwalter:
This is probably poorly written since I have no experience with kotlin.  Please let me know how I should change it (not sure where to put the new data class files, change naming etc)


MisterCheezeCake:
I think there should be some signifier that the message is coming from ChatTriggers, like maybe `&8--------- &r[&5ChatTriggers&f]&8---------` header at the top. Also I see a potential issue with the markdown of the changelog being printed into the chat. (unless there was some thing in there that I did not see that accounts for that)

camnwalter:
moved to #233 because idk how to use git properly

PR #232, opened 2021-03-29
Packet Triggers: Fix failed assertions and remove ASM

glasss13:
This commit FINALLY(hopefully) removes the issues with the packetSent/Receive triggers.

Additionally, we can now get rid of that nasty ASM and instead take advantage of the Netty pipeline to intercept packets.


Debuggingss:
Finally no more crappy ASM. Hopefully this will work (it should since I'm pretty sure this is the best and safest way to do it).

src/main/kotlin/com/chattriggers/ctjs/minecraft/listeners/ClientListener.kt
```diff
@@ -102,6 +107,31 @@ object ClientListener {
...
+                val packetReceivedEvent = CancellableEvent()
+
+                if (msg is Packet<*>)
+                    TriggerType.PACKET_RECEIVED.triggerAll(msg, packetReceivedEvent)
+
+                if (!packetReceivedEvent.isCanceled())
+                    ctx?.fireChannelRead(msg)
+            }
+
+            override fun write(ctx: ChannelHandlerContext?, msg: Any?, promise: ChannelPromise?) {
```
DJtheRedstoner, line 123: I believe this should be `channelWrite` and not `write`. While it may not make a difference, it makes the most sense to use 2 methods to match as opposed to 2 that don't (the `ctx?.write` would also need to be changed to `ctx?.fireChannelWrite`).

DJtheRedstoner (changes requested):

src/main/kotlin/com/chattriggers/ctjs/minecraft/listeners/ClientListener.kt
glasss13: I agree the naming is rather strange but there isn't any `channelWrite` method so blame netty not me.

DJtheRedstoner: Seems like that is true, I swear there was a `channelWrite` but its either a newer netty version or I am just misremembering.

glasss13:
Whats the dealio with this force push? how fix?

FalseHonesty:
> Whats the dealio with this force push? how fix?

Is this in reference to it saying matt force pushed to master? If so, you don't have to worry about it.

glasss13:
> > Whats the dealio with this force push? how fix?
> 
> Is this in reference to it saying matt force pushed to master? If so, you don't have to worry about it.

yea, ok cool.

src/main/kotlin/com/chattriggers/ctjs/minecraft/listeners/ClientListener.kt
Debuggingss: I checked netty docs and there doesn't seem to be a `channelWrite` in any of the versions. Correct me if I'm wrong. I don't think that changed in newer versions.

PR #239, opened 2021-06-29
clean up crash report category

camnwalter:
as far as I'm aware, `Module#toString` is only being used for adding the crash report section via ASMUtils, which ends up being really long with unneeded info like seen [here](https://i.imgur.com/omYFSRv.png)


mattco98:
This is a good change, but let's also include the version in the output (and also resolve the conflict of course).

PR #241, opened 2021-07-18
Improved Renderer.drawPlayer

Debuggingss:
Added 5 more arguments to Renderer.drawPlayer method (showNametag, showArmor, showCape, showHeldItem, showArrows) and a (-1, 1, 1) scale so the user no longer needs to do negative scale to use it.
The arguments are pretty self explanatory. Performance shouldn't be bad, I did some testing (Thanks to DJtheRedstoner for some performance improvements).

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/CTRenderPlayer.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/CTRenderPlayer.kt
@@ -13,11 +13,11 @@ class CTRenderPlayer(renderManager: RenderManager?, useSmallArms: Boolean) : Ren
     private var showArrows = true
 
     fun setOptions(
-            showNametag: Boolean,
-            showArmor: Boolean,
-            showCape: Boolean,
-            showHeldItem: Boolean,
-            showArrows: Boolean
+        showNametag: Boolean,
+        showArmor: Boolean,
+        showCape: Boolean,
+        showHeldItem: Boolean,
+        showArrows: Boolean
     ) {
         this.showNametag = showNametag
         this.showArmor = showArmor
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/CTRenderPlayer.kt
```diff
@@ -0,0 +1,84 @@
...
+    override fun doRender(
+            entity: AbstractClientPlayer?,
+            x: Double,
+            y: Double,
+            z: Double,
+            entityYaw: Float,
+            partialTicks: Float
+    ) {
+        super.doRender(entity, x, y, z, entityYaw, partialTicks)
+    }
```
mattco98: Unnecessary overload

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/CTRenderPlayer.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    override fun setModelVisibilities(clientPlayer: AbstractClientPlayer) {
+        super.setModelVisibilities(clientPlayer)
+        if(!showHeldItem) getMainModel().heldItemRight = 0
+    }
+
+    override fun canRenderName(entity: AbstractClientPlayer?): Boolean {
+        return showNametag
+    }
```
mattco98:
```suggestion
    override fun canRenderName(entity: AbstractClientPlayer?) = showNameTag
```

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Renderer.kt
```diff
@@ -23,6 +24,7 @@ import kotlin.math.atan2
 import kotlin.math.cos
 import kotlin.math.sin
 
+
```
mattco98: Unnecessary newline

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Renderer.kt
```diff
@@ -361,9 +363,23 @@ object Renderer {
         finishDraw()
     }
 
+    private val renderMgr = Client.getMinecraft().renderManager
```
mattco98: No need for this abbreviation, just call it `renderManager`

mattco98 (changes requested):
Please format the files as they're all missing spaces after the `if` keyword.

PR #243, opened 2021-08-09
fixed blockBreak trigger

### LINKED ISSUE #219: blockBreak trigger not activating on Servers
**Describe the bug**
When a player breaks a block while on a server, `register("blockBreak")` callback functions will not be triggered.

While playing in single player, this issue does not occur.

**To Reproduce**
1. Create a new module containing the following code:
```
register("blockBreak", (block, player, event) => {
  console.log("BREAK");
});
```
2. Call `/ct console js` to display the console.
3. Log into a single player world and break a block. See that the console displays "BREAK".
4. Log into a multiplayer world and break a block. See that the console does not display "BREAK".

**Expected behavior**
An event should be triggered, allowing the module to respond to it.

**Setup (please complete the following information):**
 - OS: Windows 10
 - ChatTriggers Version: 1.3.0
 - Minecraft Version: 1.8.9
- Java Version: 1.8.0_51 64bit

**Additional context**
I wanted to track blocks broken per second, with 20 being the maximum the game allows.

camnwalter:
now injects instead of using the server side forge event

also hitBlock trigger now returns the correct block pos instead of always returning (0, 0, 0)

~~no idea why these triggers don't work in dev, but they're working in prod~~
edit: was an idiot and didn't read the contributing.md file

fixes #219

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/launch/plugin/CTJSTransformer.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/launch/plugin/CTJSTransformer.kt
@@ -34,9 +34,7 @@ class CTJSTransformer : BaseClassTransformer() {
             injectRenderManager()
             injectEffectRenderer()
             injectGuiScreen()
-
-            // TODO: Investigate
-//            injectPlayerControllerMP()
+            injectPlayerControllerMP()
 
             ModuleManager.setup()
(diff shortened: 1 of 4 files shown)


EliteMasterEric:
@Squagward thank you very much for implementing this!

PR #249, opened 2021-09-02
fixed buttons for Gui

camnwalter:


mattco98:
Can you provide a little more details for this? It isn't really obvious what this does or why it's necessary

camnwalter:
> Can you provide a little more details for this? It isn't really obvious what this does or why it's necessary


Buttons weren't being created previously when you used addButton, so I looked up other examples of adding GuiButton and they added them inside the initGui method

PR #256, opened 2021-11-13
Implement String.prototype.replaceAll

MutantPiggieGolem1:
Implementation should be standard with vanilla js functionality.


mattco98:
This was said somewhere a while ago, but I think this change would be better made in Rhino itself.

mattco98:
Implemented in ChatTriggers/rhino@60f40a81851d6f234f83f45ff2c68eda5d5c44ad

PR #257, opened 2021-11-14
ChatLib: Add addToSentMessageHistory

### LINKED ISSUE #254: Add a way to insert text into the player's chat history buffer
"Chat history buffer" being the thing you cycle through when you hit the up/down arrow keys with the chat GUI open.

camnwalter:
Not sure if it should be called `addToChatHistory` or something like `addToSentMessages`. The only thing I can think of that makes it confusing is `ChatLib#getChatLines` retrieves `ChatListener.chatHistory`, which are the actual messages in chat, not the player's sent messages.

closes #254

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/ChatLib.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/ChatLib.kt
@@ -58,7 +58,7 @@ object ChatLib {
      * Simulates a chat message to be caught by other triggers for testing.
      * The text can be a String, a [Message] or a [TextComponent]
      *
-     * @param message The message to simulate
+     * @param text The message to simulate
      */
     @JvmStatic
     fun simulateChat(text: Any) {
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/ChatLib.kt
```diff
@@ -326,6 +322,17 @@ object ChatLib {
...
+    /**
+     * Adds a message to the player's chat history. This allows the message to
+     * show up for the player when pressing the up/down keys while in the chat gui
+     *
+     * @param message the message to add to chat history
+     */
+    @JvmStatic
+    fun addToChatHistory(message: String) {
+        Client.getChatGUI()!!.addToSentMessages(message)
+    }
```
mattco98: How about we add an index parameter for a bit more flexibility
```suggestion
    @JvmStatic
    @JvmOverloads
    fun addToChatHistory(message: String, index: Int? = null) {
        val sentMessages = Client.getChatGUI()!!.getSentMessages()
        sentMessages.add(index ?: sentMessages.size, message)
    }
```
Also, how about `addToSentMessageHistory`, that seem a bit less ambiguous.

mattco98 (changes requested):

PR #259, opened 2021-11-24
Add RenderSlotHighlight & PreItemRender triggers, Move Settings into Client, Change PostGuiRender param order

camnwalter:
- RenderSlotHighlight is fired whenever the mouse is hovering over a slot in the inventory and can cancel the white square from showing up.
- PreItemRender is a way to render after the gui background is drawn, but before the items are drawn. Here's an example of it in action: 
image

- Moved Settings object into Client, this seems to be an issue with kotlin not liking triple nested objects being split up.  The issue can be seen [here](https://pl.kotl.in/L6tJfvid1)

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/launch/plugin/guiContainer.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/launch/plugin/guiContainer.kt
@@ -1,0 +1,101 @@
+package com.chattriggers.ctjs.launch.plugin
+
+import com.chattriggers.ctjs.minecraft.listeners.CancellableEvent
+import com.chattriggers.ctjs.triggers.TriggerType
+import dev.falsehonesty.asmhelper.dsl.At
+import dev.falsehonesty.asmhelper.dsl.InjectionPoint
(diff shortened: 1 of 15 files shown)


camnwalter:
Please let me know if I should pass in other parameters to the triggers.  I'll change top and left gui coords to top and left coords of the slot, but is there anything else should I add

src/main/kotlin/com/chattriggers/ctjs/launch/plugin/guiContainer.kt
```diff
@@ -0,0 +1,102 @@
...
+        code {
+            if (theSlot != null) {
+                val event = CancellableEvent()
+
+                TriggerType.RenderSlotHighlight.triggerAll(local1, local2, theSlot, local0, event)
+
+                if (event.isCancelled()) {
+                    asm {
+                        pop()
+                        pop()
```
mattco98: Nitpick: `pop2`

mattco98 (changes requested):
I think making `Settings` a class instead of an object is a better solution. It's not exposed directly to JS as a binding anyways.

Also, the first two commits can be squashed.

PR #260, opened 2021-11-26
Gui: Add a bunch of methods for GuiButtons

camnwalter:


src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/gui/Gui.kt
```diff
@@ -241,33 +241,130 @@ abstract class Gui : GuiScreen() {
...
+    fun getButtonVisibility(buttonId: Int): Boolean = getButton(buttonId)?.visible ?: false
+
+    /**
+     * Sets the visibility of a button
+     *
+     * @param buttonId the id of the button to change
+     * @param visible the new visibility of the button
+     * @return the GuiButton for method chaining
+     */
+    fun setButtonVisibility(buttonId: Int, visible: Boolean) = getButton(buttonId)?.apply {
```
mattco98: This should return the `Gui` rather than the button itself. Applies for all of the setters below.

mattco98 (changes requested):

mattco98 (approved):

PR #269, opened 2021-12-18
Added PostRenderEntity trigger and /ct reload shortcut to console

Debuggingss:
Added a postRenderEntity trigger so you can actually reset transformations done on renderEntity.
Added Control + R as a shortcut to JavaScript console to reload ChatTriggers.
Added 'help' keyword to console to print out all shortcuts.


Debuggingss:
There was a miniscule amount of tomfoolery going on. Sorry about that.

PR #279, opened 2021-12-31
Fixed Message constructor & other various changes

camnwalter:
- Fixed Message constructor only creating 1 Text Component when using the MCITextComponent constructor
- Added support for drawing shapes with clockwise vertices
- Changed KeyBind's constructor to throw an error if a KeyBind already exists
- Slightly refactored Text

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Shape.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Shape.kt
@@ -12,9 +12,11 @@ import kotlin.math.cos
 import kotlin.math.sin
 
 @External
-class Shape(private var color: Int) {
-    private var vertexes = mutableListOf<Vector2f>()
+class Shape(private var color: Long) {
+    private val vertexes = mutableListOf<Vector2f>()
+    private val reversedVertexes = vertexes.asReversed()
     private var drawMode = 9
+    private var area = 0f
 
     fun copy(): Shape = clone()
 
@@ -25,9 +27,9 @@ class Shape(private var color: Int) {
         return clone
     }
 
-    fun getColor(): Int = color
+    fun getColor(): Long = color
 
-    fun setColor(color: Int) = apply { this.color = color }
+    fun setColor(color: Long) = apply { this.color = Renderer.fixAlpha(color) }
 
     fun getDrawMode(): Int = drawMode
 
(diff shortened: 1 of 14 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/message/Message.kt
```diff
@@ -34,7 +35,27 @@ class Message {
      * @param component the IChatComponent
      */
     constructor(component: MCITextComponent) {
-        messageParts.add(TextComponent(component))
+        if (component.siblings.isEmpty()) {
```
mattco98, line 38: Yeah I'm just gonna trust this works cause I have no clue whats going on here

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Shape.kt
```diff
@@ -100,24 +100,22 @@ class Shape(private var color: Int) {
...
-        if (Renderer.colorized == null)
-            GlStateManager.color(r, g, b, a)
+        Renderer.doColor(color.toLong())
 
         worldRenderer.begin(drawMode, DefaultVertexFormats.POSITION)
 
-        for (vertex in vertexes)
-            worldRenderer.pos(vertex.x.toDouble(), vertex.y.toDouble(), 0.0).endVertex()
+        if (!Renderer.isCounterClockwise(vertexes))
+            vertexes.reverse()
```
mattco98: I think it would be better to store a potentially-reversed view of the list instead of doing this check every frame. There might be a reversed list view class somewhere in the stdlib, but if not it should be easy enough to make one. It would have to be rechecked every time the vertex list is mutated.

src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/KeyBind.kt
```diff
@@ -22,24 +21,27 @@ class KeyBind {
...
-        for (key in Client.getMinecraft().gameSettings.keyBindings) {
-            if (key.keyCategory == category && key.keyDescription == description) {
-                Client.getMinecraft().gameSettings.keyBindings = ArrayUtils.removeElement(
-                    Client.getMinecraft().gameSettings.keyBindings, key
-                )
-                break
-            }
+        val possibleDuplicate = Client.getMinecraft().gameSettings.keyBindings.find {
+            I18n.format(it.keyDescription) == I18n.format(description) &&
+            I18n.format(it.keyCategory) == I18n.format(category)
```
mattco98, line 26: Is there a reason to format these? If the translation key is the same, the translation would also be the same, right?

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Text.kt
```diff
@@ -77,52 +81,45 @@ class Text @JvmOverloads constructor(private var string: String, private var x:
...
-            Renderer.getFontRenderer().drawString(
-                string,
-                getXAlign(string, x ?: this.x),
-                (y ?: this.y) / scale,
-                color.toInt(),
-                shadow
-            )
+
+        val longestLine = lines.maxOf { Renderer.getStringWidth(it) * scale }
+        if (maxWidth != 0) longestLine.coerceAtMost(maxWidth.toFloat())
```
mattco98: Missing assignment

mattco98 (changes requested):

src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/KeyBind.kt
camnwalter: I just tested it, and these are the results of not comparing the description and categories formatted.

- Attack/Destroy, Gameplay => game crashes from error
- Attack/Destroy, key.categories.gameplay => creates a "duplicate" keybind (appears to have same name and category, but is completely separate)
- key.attack, key.categories.gameplay => game crashes from error
- key.attack, Gameplay => game crashes from error

mattco98: lit

PR #284, opened 2022-01-28
Added bunch of triggers for better keybind handling

Debuggingss:
KeyPress and KeyDown will fire for ALL keys and they return the keycode of the pressed key.
KeyBindPress and KeyBindDown will fire only for ct KeyBinds. You have to use the newly added '.equals' method to compare the returned key with your KeyBind, or you can use the 2nd argument which is the keycode, again this is only for ct KeyBinds.

(note: none of these triggers will fire inside GUIs)

Example usage:
```js
let bind1 = new KeyBind("Bind 1", Keyboard.KEY_G);
let bind2 = new KeyBind("Bind 2", Keyboard.KEY_H);

register('keyBindPress', (key, keycode) => {
	print(keycode);
	if (key.equals(bind1)) print("Bind 1 has been pressed");
	else if (key.equals(bind2)) print("Bind 2 has been pressed");
});

register('keyBindDown', (key, keycode) => {
	print(keycode);
	if (key.equals(bind1)) print("Bind 1 is held down");
	else if (key.equals(bind2)) print("Bind 2 is held down");
});

register('keyPress', (keycode) => {
	print("Key Pressed: " + keycode);
});

register('keyDown', (keycode) => {
	print("Key Down: " + keycode);
});
```

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/engine/IRegister.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/engine/IRegister.kt
@@ -794,6 +794,52 @@ interface IRegister {
     }
 
     /**
+     * Registers a new trigger that runs when a [KeyBind] is pressed
+     *
+     * Passes through two arguments:
+     * - The [KeyBind] that has been pressed
+     * - The keycode of the KeyBind
+     *
+     */
+    fun registerKeyBindPress(method: Any): OnRegularTrigger {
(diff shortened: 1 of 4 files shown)


Debuggingss:
Closing pull request as @Squagward made a much better implementation. [Squagward's PR](https://github.com/ChatTriggers/ChatTriggers/pull/286)

PR #285, opened 2022-01-30
Config: Sort General before Console category, guiContainer: Fix triggers using incorrect obfuscated names

camnwalter:


src/main/kotlin/com/chattriggers/ctjs/launch/plugin/types.kt
```diff
@@ -18,3 +18,4 @@ const val BLOCK_POS = "net/minecraft/util/BlockPos"
 const val ENUM_FACING = "net/minecraft/util/EnumFacing"
 const val TRIGGER_TYPE = "com/chattriggers/ctjs/triggers/TriggerType"
 const val TILE_ENTITY = "net/minecraft/tileentity/TileEntity"
+const val GUICONTAINER = "net/minecraft/client/gui/inventory/GuiContainer"
```
mattco98:
```suggestion
const val GUI_CONTAINER = "net/minecraft/client/gui/inventory/GuiContainer"
```

src/main/kotlin/com/chattriggers/ctjs/utils/CategorySorting.kt
```diff
@@ -0,0 +1,13 @@
+package com.chattriggers.ctjs.utils
+
+import gg.essential.vigilance.data.Category
+import gg.essential.vigilance.data.SortingBehavior
+
+object CategorySorting: SortingBehavior() {
```
mattco98: Missing a space

src/main/kotlin/com/chattriggers/ctjs/utils/CategorySorting.kt
```diff
@@ -0,0 +1,13 @@
...
+    override fun getCategoryComparator(): Comparator<in Category> {
```
mattco98, line 7: Lets make this throw if a category isn't in the list, that way if another category gets added in the future the implementer won't forget to add it here.

src/main/kotlin/com/chattriggers/ctjs/utils/CategorySorting.kt
```diff
@@ -0,0 +1,13 @@
...
+        return Comparator { o1, o2 ->
+            val categories = listOf("General", "Console")
+            return@Comparator categories.indexOf(o1.name) - categories.indexOf(o2.name)
```
mattco98: Useless `return`

mattco98 (changes requested):

PR #286, opened 2022-02-03
KeyBind: Add registerKey{Press/Release/Down} methods

camnwalter:
Moved from #284


mattco98:
3a4240a7965d54c5ac466d2a2ae02db47fa60753 appears to be breaking

camnwalter:
> [3a4240a](https://github.com/ChatTriggers/ChatTriggers/commit/3a4240a7965d54c5ac466d2a2ae02db47fa60753) appears to be breaking

Should I just remove the lookingAt changes and keep the rest?  I can make an issue to add it whenever the next best time would be.  I just added it since it is annoying with the method returning BlockType when looking at air and Block when looking at blocks.

mattco98:
Yeah that'd probably be best

PR #291, opened 2022-02-25
Display and DisplayHandler changes

DalwynWasTaken:
- Made addLine not return void. This allows method chaining
- Added the ability to change the event on which the display is drawn.
    - Built in registers: renderOverlay (default) and postGuiRender
    - Ability to change it, so you can draw the display manually

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/display/DisplayHandler.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/display/DisplayHandler.kt
@@ -3,6 +3,7 @@ package com.chattriggers.ctjs.minecraft.objects.display
 import com.chattriggers.ctjs.utils.kotlin.External
 import net.minecraft.client.renderer.GlStateManager
 import net.minecraftforge.client.event.RenderGameOverlayEvent
+import net.minecraftforge.client.event.GuiScreenEvent
 import net.minecraftforge.common.MinecraftForge
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
 import java.util.concurrent.CopyOnWriteArrayList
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/display/DisplayHandler.kt
```diff
@@ -20,12 +21,31 @@ object DisplayHandler {
...
+        displays.forEach {
+            if(it.registerType == RegisterType.POSTGUIRENDER) {
+                it.render()
+            }
+        }
+        GlStateManager.popMatrix()
+    }
+
+    enum class RegisterType {
+        RENDEROVERLAY, POSTGUIRENDER, OTHER
```
mattco98: `OTHER` doesn't seem necessary, Displays aren't necessary supposed to be insanely customizable, so let's just add the post gui render thing here.

mattco98 (changes requested):

src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/display/DisplayHandler.kt
camnwalter: Also I think it would be better to use SCREAMING_SNAKE_CASE to be consistent with DisplayHandler.Background. (RENDER_OVERLAY etc)

DalwynWasTaken: OTHER has been removed and the Enum classes have been made consistent with DisplayHandler's other Enum classes. The files have been changed to accommodate this change and it's been rebased to a single commit

src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/display/DisplayHandler.kt
```diff
@@ -20,12 +21,31 @@ object DisplayHandler {
...
+            }
+        }
         GlStateManager.popMatrix()
     }
 
+    @SubscribeEvent
+    fun renderDisplayGui(event: GuiScreenEvent.DrawScreenEvent.Post) {
+        GlStateManager.pushMatrix()
+        displays.forEach {
+            if(it.registerType == RegisterType.POST_GUI_RENDER) {
```
camnwalter: There should be a space after if, same with the one above

DalwynWasTaken: Alright, I assume that's all the spelling checks you have correct?

PR #292, opened 2022-02-27
Various bugfixes

camnwalter:


src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/display/DisplayLine.kt
```diff
@@ -53,12 +57,17 @@ abstract class DisplayLine {
 
     init {
         MouseListener.registerClickListener { x, y, button, pressed ->
-            if (x in cachedX..cachedX + cachedWidth && y in cachedY..cachedY + cachedHeight)
+            if (
+                shouldRender &&
+                x in cachedX..cachedX + cachedWidth &&
+                y in cachedY..cachedY + cachedHeight
+            )
                 onClicked?.trigger(arrayOf(x, y, button, pressed))
```
mattco98, line 65: I like to put braces around `if` statements with a multiline condition, even if they're only one line

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Player.kt
```diff
@@ -34,6 +35,11 @@ object Player {
         //#endif
     }
 
+    @JvmStatic
+    fun getTeam(): ScorePlayerTeam? {
```
mattco98: Seems like this should have a wrapper if we're adding a utility method to retrieve it.

mattco98 (changes requested):

PR #301, opened 2022-03-28
ModuleManager: Allow modules with jars to be deleted via /ct delete command

camnwalter:
This fixes the issue of having users have to close the game and delete the folder manually to delete modules that had jars inside.

Thanks to @DJtheRedstoner who mentioned how to do this a bit ago


DJtheRedstoner (approved):
LGTM. Refactoring JAR loading entirely would be good at some point but this is a reasonable solution in the meantime.

PR #305, opened 2022-04-02
Add draw slot trigger

appable0:
Hi – this adds a trigger that runs as slots in containers are being drawn, allowing item rendering to be cancelled within containers (e.g. for removing background items in containers used as GUIs or visually replacing an item). First pull request to anything so apologies if this is very wrong.


mattco98:
It'd be nice to add a small `Slot` wrapper to go along with this so that people don't need to use the obfuscated names when using this. Doesn't have to be everything, just the parts you feel like would be the most used. Other than that this looks good though, thanks for the contribution!

appable0:
Added. I kept the current triggers that pass MC slots as is for compatibility – not sure if that's best?

PR #306, opened 2022-04-10
Add additional item drawing features

appable0:
- Adds drawWithOverlay() method for rendering items with their stack size and damage, as in vanilla (basically emulating renderItemOverlayInGui)
- Adds optional arguments for z-level, custom overlay text (e.g. stack size), and custom "durability bar" in draw().
- Change setStackSize() to not remove damage / other ItemStack data (may break modules that relied on this, but seems unlikely anyone would rely on this).


appable0:
Going to merge this and other into one more general pull request for item/slot rendering changes.

PR #307, opened 2022-04-14
Item rendering triggers

appable0:
Some new triggers and utilities around rendering items and slots:
- renderSlot(Slot, gui, event) triggers when a slot is being drawn
- renderItemIntoGui(Item, x, y, event) triggers for each item being drawn into a gui (doesn't trigger on Item.draw)
- renderItemOverlayIntoGui(Item, x, y, event) triggers for each item overlay being drawn into a GUI.
- Simple slot wrapper
- Client.currentGui.getSlotUnderMouse() method.
- Add z-level optional argument to Item.draw(x, y, scale, z) for positioning.

(the renderItemIntoGui and renderItemOverlayIntoGui duplication is a bit inelegant, but I don't think there's many times where you would actually want to cancel both and an impractical number of methods call those two)

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/engine/IRegister.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/engine/IRegister.kt
@@ -2,6 +2,7 @@ package com.chattriggers.ctjs.engine
 
 import com.chattriggers.ctjs.minecraft.listeners.ClientListener
 import com.chattriggers.ctjs.minecraft.wrappers.objects.inventory.Item
+import com.chattriggers.ctjs.minecraft.wrappers.objects.inventory.Slot
 import com.chattriggers.ctjs.triggers.*
 import com.chattriggers.ctjs.utils.kotlin.External
 import kotlin.reflect.KFunction
@@ -1196,7 +1197,7 @@ interface IRegister {
      * Passes through five arguments:
      * - The mouseX position
      * - The mouseY position
-     * - The Slot
+     * - The MC Slot
      * - The GuiContainer
      *
      * Available modifications:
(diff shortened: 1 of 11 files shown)


src/main/kotlin/com/chattriggers/ctjs/launch/plugin/guiContainer.kt
```diff
@@ -1,21 +1,58 @@
...
+    methodName = "drawSlot"
+    methodDesc = "(Lnet/minecraft/inventory/Slot;)V"
+
+    at = At(InjectionPoint.HEAD)
+
+    methodMaps = mapOf(
+        "func_146977_a" to "drawSlot",
+    )
+
+    fieldMaps = mapOf("theSlot" to "field_147006_u")
```
camnwalter: You can remove this since it's not used.
```suggestion

```

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/objects/inventory/Slot.kt
```diff
@@ -0,0 +1,21 @@
+package com.chattriggers.ctjs.minecraft.wrappers.objects.inventory
+
+import com.chattriggers.ctjs.utils.kotlin.External
+import com.chattriggers.ctjs.utils.kotlin.MCSlot
+import net.minecraft.client.gui.inventory.GuiContainer
+
+@External
+class Slot(val mcSlot: MCSlot) {
```
camnwalter, line 7: Can you also add Slot to moduleProvidedLibs?

Squagward, line 8: I think Slot should also be added to the moduleProvidedLibs file.

camnwalter (changes requested):
Thanks for the contribution!  Looks pretty good overall, but I just have a couple of very tiny nitpicks in the following comments.

appable0:
Fixed both (I think), thanks!

src/main/kotlin/com/chattriggers/ctjs/engine/IRegister.kt
```diff
@@ -1209,13 +1210,62 @@ interface IRegister {
...
+
+    /**
+     * Registers a new trigger that runs before each item overlay (stack size and damage bar) is drawn.
+     *
+     * Passes through four arguments:
+     * - The [Item]
+     * - The x position
+     * - The y position
+     * - The event, which can be cancelled.
+
```
mattco98: Remove newline

src/main/kotlin/com/chattriggers/ctjs/triggers/TriggerType.kt
```diff
@@ -60,6 +61,8 @@ enum class TriggerType {
     RenderEntity,
     PostGuiRender,
     PreItemRender,
+    RenderItemIntoGUI,
+    RenderItemOverlayIntoGUI,
```
mattco98: `GUI` -> `Gui` (for both triggers)

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/objects/inventory/Slot.kt
```diff
@@ -0,0 +1,21 @@
...
+
```
mattco98: Remove newline

mattco98 (changes requested):
LGTM, just a few nitpicks. Also please squash your cleanup commits into the main commit.

PR #308, opened 2022-04-17
Add colors to console

camnwalter:
This adds the option for colors in the console via `console.log`, `console.warn`, `console.error`, and `console.assert`.  Also added an option to `print` to take in a second argument allowing for custom colors.  I didn't add that to `console.log` or its other methods since they already take varargs, making it impossible to tell if the user wanted to print the color in string form, or wanted to color the text.

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/utils/Config.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/utils/Config.kt
@@ -122,4 +122,37 @@ object Config : Vigilant(File(CTJS.configLocation, "ChatTriggers.toml"), sorting
         category = "Console",
     )
     var consoleBackgroundColor = Color(21, 21, 21)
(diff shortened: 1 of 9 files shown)


src/main/kotlin/com/chattriggers/ctjs/utils/Config.kt
```diff
@@ -122,4 +122,39 @@ object Config : Vigilant(File(CTJS.configLocation, "ChatTriggers.toml"), sorting
...
+
+    @Property(
+        PropertyType.COLOR,
+        name = "Console warning color",
+        category = "Console",
+    )
+    var consoleWarningColor = Color(248, 191, 84)
+
+    init {
+        val clazz = javaClass
```
mattco98: Is this temporary necessary?

PR #312, opened 2022-04-24
Item: Add setName and setLore methods

camnwalter:
Also fixes getLore returning the item's name as the first element in the lore list.

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/inventory/Item.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/inventory/Item.kt
@@ -1,20 +1,21 @@
 package com.chattriggers.ctjs.minecraft.wrappers.inventory
 
+import com.chattriggers.ctjs.minecraft.libs.ChatLib
 import com.chattriggers.ctjs.minecraft.libs.renderer.Renderer
 import com.chattriggers.ctjs.minecraft.objects.message.TextComponent
 import com.chattriggers.ctjs.minecraft.wrappers.Client
 import com.chattriggers.ctjs.minecraft.wrappers.Player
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/inventory/Item.kt
```diff
@@ -181,12 +193,39 @@ class Item {
 
     fun getLore(): List<String> {
         //#if MC<=10809
-        return itemStack.getTooltip(Player.getPlayer(), Client.getMinecraft().gameSettings.advancedItemTooltips)
+        return itemStack.getTooltip(Player.getPlayer(), Client.getMinecraft().gameSettings.advancedItemTooltips).drop(1)
```
mattco98: This seems very breaking

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/inventory/Item.kt
```diff
@@ -127,7 +131,15 @@ class Item {
      *
      * @return the item's stack display name
      */
-    fun getName(): String = if (getID() == 0) "air" else itemStack.displayName
+    fun getName(): String = itemStack.displayName
```
mattco98, line 134: Is this breaking as well? What is air's `displayName`?

camnwalter: Air is technically null, so an Item can't be made out of air since Item not take null Item stacks.  The check was useless when this was changed.  I added the checks to the constructors since if you tried to do new Item(0) it would cause an internal error, since that would make the Item stack null.

PR #318, opened 2022-04-27
Slot: Correctly number the slot indices inside GuiContainerCreative

camnwalter:
Not sure if there's a better way to check the class since ContainerCreative is package private.


mattco98:
You should just be able to make it public with ATs right?

PR #323, opened 2022-04-29
KeyBind: Re-add clearing & removing keybinds, …

camnwalter:
I also changed to make Client.getKeyBindFrom... methods attempt to return an existing KeyBind before making a new one.  Not sure if this is considered breaking or not but wanted to include it in case it wasn't.

Like my comment says in KeyBindHandler line 34, there were occasional NPE's caused by this, but catching during tick fixes this.  I'm not sure if there's anything we can do about it, since it happened too randomly for me to notice a pattern, and the game works perfectly normal with it.

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/keybind/KeyBindHandler.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/keybind/KeyBindHandler.kt
@@ -4,25 +4,39 @@ import com.chattriggers.ctjs.minecraft.wrappers.World
 import net.minecraftforge.common.MinecraftForge
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
 import net.minecraftforge.fml.common.gameevent.TickEvent
+import java.util.concurrent.CopyOnWriteArrayList
 
 internal object KeyBindHandler {
     init {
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/keybind/KeyBindHandler.kt
```diff
@@ -4,25 +4,39 @@ import com.chattriggers.ctjs.minecraft.wrappers.World
...
         keyBinds.add(keyBind)
     }
 
+    fun unregisterKeyBind(keyBind: KeyBind) {
+        keyBinds.remove(keyBind)
+    }
+
     fun clearKeyBinds() = keyBinds.clear()
 
+    fun getKeyBinds() = keyBinds
```
mattco98, line 26: This can probably be internal since JS shouldn't need this

camnwalter: The whole object is internal already.

mattco98: you right you right

PR #325, opened 2022-05-04
Make OnRenderTrigger more generic; Remove "On" trigger prefix

mattco98:


glasss13:
Since `OnTrigger` is now called `Trigger` it seems reasonable to rename all the other triggers similarly to keep it consistent

PR #326, opened 2022-05-05
Misc: Wrap tasks needing Minecraft's main thread inside scheduleTask

camnwalter:
There might be more methods that need the main thread that I missed, but these were what I could find.

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/listeners/ClientListener.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/listeners/ClientListener.kt
@@ -3,11 +3,12 @@ package com.chattriggers.ctjs.minecraft.listeners
 import com.chattriggers.ctjs.engine.langs.js.JSContextFactory
 import com.chattriggers.ctjs.minecraft.libs.ChatLib
 import com.chattriggers.ctjs.minecraft.libs.EventLib
+import com.chattriggers.ctjs.minecraft.wrappers.Client
 import com.chattriggers.ctjs.minecraft.wrappers.Scoreboard
 import com.chattriggers.ctjs.minecraft.wrappers.World
-import com.chattriggers.ctjs.minecraft.wrappers.world.block.BlockFace
 import com.chattriggers.ctjs.minecraft.wrappers.entity.PlayerMP
 import com.chattriggers.ctjs.minecraft.wrappers.inventory.Item
+import com.chattriggers.ctjs.minecraft.wrappers.world.block.BlockFace
 import com.chattriggers.ctjs.printToConsole
 import com.chattriggers.ctjs.triggers.TriggerType
 import com.chattriggers.ctjs.utils.Config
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/listeners/ClientListener.kt
```diff
@@ -67,7 +69,18 @@ object ClientListener {
...
-        if (World.getWorld() == null || event.phase == TickEvent.Phase.END)
+        if (event.phase == TickEvent.Phase.END)
+            return
+
+        tasks.removeAll {
+            if (it.delay-- <= 0) {
+                it.callback()
+                return@removeAll true
+            }
+            false
```
mattco98: Small code cleanliness nitpick
```suggestion
            if (it.delay-- <= 0) {
                it.callback()
                true
            } else false
```

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Client.kt
```diff
@@ -82,19 +85,39 @@ abstract class Client {
...
+         * @param callback The task to run on the main thread
+         */
+        @JvmOverloads
+        @JvmStatic
+        fun scheduleTask(delay: Int = 0, callback: () -> Unit) {
+            ClientListener.tasks.add(Task(delay) {
+                if (getMinecraft().isCallingFromMinecraftThread) {
+                    callback()
+                } else {
+                    getMinecraft().addScheduledTask(callback)
```
mattco98: Is this if statement necessary? It looks like `Minecraft::addScheduledTask` already includes a check for `isCallingFromMinecraftThread` which leads to directly calling the `Runnable` if it is `true`.

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Client.kt
```diff
@@ -57,6 +58,8 @@ abstract class Client {
      */
     abstract fun getKeyBindFromDescription(description: String): KeyBind?
 
+    class Task(var delay: Int, val callback: () -> Unit)
```
mattco98: Should probably reside in `ClientListener`

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Client.kt
mattco98: Also, if `Task` is moved to `ClientListener`, it can probably just hold `callback`, and `ClientListener::onTick` can handle the scheduling stuff

src/main/kotlin/com/chattriggers/ctjs/minecraft/listeners/ClientListener.kt
```diff
@@ -33,6 +34,7 @@ object ClientListener {
     private var ticksPassed: Int = 0
     val chatHistory = mutableListOf<String>()
     val actionBarHistory = mutableListOf<String>()
+    val tasks = mutableListOf<Task>()
```
mattco98: This either needs to be a concurrent data structure, or have synchronized accesses at the use sites

mattco98 (changes requested):

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Client.kt
camnwalter: > Also, if `Task` is moved to `ClientListener`, it can probably just hold `callback`, and `ClientListener::onTick` can handle the scheduling stuff

What do you mean by this? I had it so you could schedule something x ticks in the future, so how would that work without storing the delay with it?

mattco98: Sorry, I meant hold `callback` instead of the wrapping lambda. It'd still hold delay, so this code would just be `ClientListener.tasks.add(Task(delay, callback))`

PR #329, opened 2022-05-07
Add author field to module metadata

Debuggingss:
Since a few people use author (including me) I feel like this is a useful addition. The author/creator field in the metadata is not actually checked anywhere.


mattco98:
The creator field is used by the frontend, so all uses of `author` should be changed to `creator` instead of adding it as a new field

Debuggingss:
Closing as it's useless. Just use `creator`.

PR #333, opened 2022-05-19
ClientListener: Fix "high ping" lag with packets

camnwalter:
Previously, playing on certain servers (e.g. Hypixel's housing lobby), players would have to wait sometimes 30 seconds or longer to move around and open menus without being rubberbanded back.  Somehow this was due to the packet "fix" I added a couple of months ago, but I'm still not sure why it would cause so much lag like that.  I'm not really sure why that caused so much lag, but it might have been an issue with Rhino attempting to create a context for every packet sent and received, but that doesn't make sense since it should store and reuse contexts for the same thread.


mattco98:
If it works it works

PR #334, opened 2022-06-12
Multiversion: Add basic Fabric support

camnwalter:
This allows the game to launch but there are still tons of things that need to be tested.  It's a good start.


src/main/kotlin/com/chattriggers/ctjs/utils/Config.kt
```diff
@@ -8,14 +8,6 @@ import java.awt.Color
 import java.io.File
 
 object Config : Vigilant(File(CTJS.configLocation, "ChatTriggers.toml"), sortingBehavior = CategorySorting) {
-    @Property(
```
mattco98, line 11: Let's document this change with a `TODO(BREAKING)`

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
```diff
@@ -51,59 +63,59 @@ import net.minecraftforge.fml.common.event.FMLInitializationEvent
...
-            thread {
-                try {
-                    ModuleManager.setup()
-                    ModuleManager.entryPass()
+            ModuleUpdater
+            UpdateChecker
+            MouseListener
+            CPS
+            WorldListener
+            ClientListener
```
mattco98: Looks like something was forgotten here

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/ChatLib.kt
```diff
@@ -193,13 +189,9 @@ object ChatLib {
...
         val spaceWidth = (chatWidth - textWidth) / 2f
-        val spaceBuilder = StringBuilder().apply {
-            repeat((spaceWidth / Renderer.getStringWidth(" ")).roundToInt()) {
-                append(' ')
-            }
-        }
+        val spaceBuilder = " " * (spaceWidth / Renderer.getStringWidth(" ")).roundToInt()
 
-        return spaceBuilder.append(text).toString()
+        return spaceBuilder + text
```
mattco98: This can be inlined

src/main/kotlin/com/chattriggers/ctjs/engine/module/ModuleManager.kt
```diff
@@ -19,12 +19,11 @@ import java.net.URLClassLoader
...
-    val generalConsole = Console(null)
-    val cachedModules = mutableListOf<Module>()
-    val modulesFolder = run {
-        Config.loadData()
-        File(Config.modulesFolder)
+    val generalConsole by lazy {
+        Console(null)
     }
+    val cachedModules = mutableListOf<Module>()
+    val modulesFolder = File("./config/ChatTriggers/modules")
```
mattco98: Let's make this a constant in `CTJS`, and use it here and in `FileLib`

src/main/resources/js/providedLibs.js
```diff
@@ -10,26 +10,31 @@
...
-        throw new Error(`moduleProvidedLibs: Could not load class "${path}"`);
+        if (!failSilently)
+            throw new Error(`moduleProvidedLibs: Could not load class "${path}"`);
     }
 
     function loadClass(path, className = getClassName(path)) {
         global[className] = getClass(path);
     }
 
+    function loadPossibleClass(path, className = getClassName(path)) {
```
mattco98: The functionality in `providedLibs.js` should be the exact same across versions to discourage version-specific user code, so I'm not a fan of this. If a class doesn't exist in all versions, we should make a wrapper for it. Since this PR is already huge, I'd suggest reverting these changes and just commenting out the `loadClass` calls which can fail, and we can address it in the future.

mattco98 (changes requested):

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
camnwalter: oops

src/main/resources/js/providedLibs.js
```diff
@@ -10,11 +10,12 @@
 
     const getClassName = path => path.substring(path.lastIndexOf('.') + 1)
 
-    function getClass(path) {
+    function getClass(path, failSilently = false) {
```
mattco98: Doesn't look like this is used

mattco98 (changes requested):

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
mattco98: This doesn't look fixed

camnwalter: What is wrong there?

mattco98: I guess I see what's happening. Listing classes like that with no obvious side effects just looks like a bug. It's fine for now.

PR #338, opened 2022-07-02
Verified and converted parts

camnwalter:


build.gradle.kts
```diff
@@ -22,7 +23,10 @@ loom {
     if (isForge) {
         forge {
             accessTransformer(rootProject.file("src/main/resources/$accessTransformerName"))
+            mixinConfigs.set(mutableListOf("mixinName"))
```
mattco98: This looks like it shouldn't be a string (though it was before so idk)

src/main/kotlin/com/chattriggers/ctjs/commands/Command.kt
```diff
@@ -1,119 +1,51 @@
...
-//$$ import net.minecraft.commands.Commands
-//$$ import com.mojang.brigadier.tree.CommandNode
-//#endif
+import gg.essential.api.EssentialAPI
+import gg.essential.api.commands.DefaultHandler
+import gg.essential.api.commands.Greedy
+import gg.essential.api.commands.Command as EssentialCommand
 
+// TODO(BREAKING): Removed callback parameter since nested subcommands are very difficult to implement,
+// you have to use a subcommand and then @Options
```
mattco98, line 10: I definitely don't love this. Would be nice to make a feature suggestion in the Essential discord for dynamic tab completion support so we can keep this around (if you haven't already done that).

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/CTRenderPlayer.kt
```diff
@@ -100,11 +108,12 @@ internal class CTRenderPlayer(renderManager: RenderManager, useSmallArms: Boolea
...
-        //$$ // TODO(CONVERT): Hide these behind flags?
-        //$$ addLayer(ParrotOnShoulderLayer(this, context.modelSet))
-        //$$ addLayer(SpinAttackEffectLayer(this, context.modelSet))
-        //$$ addLayer(BeeStingerLayer(this))
+        //$$ if (showParrot)
+        //$$     addLayer(ParrotOnShoulderLayer(this, context.modelSet))
+        //$$ if (showSpinAttackEffect)
+        //$$     addLayer(SpinAttackEffectLayer(this, context.modelSet))
+        //$$ if (showBeeStinger)
+        //$$     addLayer(BeeStingerLayer(this))
```
mattco98, line 121: These flags should be added to the method signature with default params

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Player.kt
```diff
@@ -383,7 +383,7 @@ object Player {
         //#if MC<=11202
         return getPlayer()?.openContainer?.let(::Inventory)
         //#else
-        //$$ return getPlayer()?.inventory?.let(::Inventory)
+        //$$ return getPlayer()?.containerMenu?.let(::Inventory)
```
mattco98, line 386: Is this breaking?

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/entity/Entity.kt
```diff
@@ -392,25 +375,26 @@ open class Entity(val entity: net.minecraft.entity.Entity) {
...
     }
 
-    fun setAngles(yaw: Float, pitch: Float) = apply {
+    fun getDimension(): World.Dimension {
         //#if MC<=11202
-        entity.setAngles(yaw, pitch)
-        //#else
-        //$$ entity.xRot = yaw
-        //$$ entity.yRot = pitch
+        return World.Dimension.values()[entity.dimension + 1]
```
mattco98, line 390: Is this guaranteed to be safe? For example, on the title screen. If not, it might be better to do `.getOrElse(entity.dimension + 1, World.Dimension.Overworld)`.

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Settings.kt
```diff
@@ -164,15 +165,18 @@ class Settings {
         fun setAmbient(level: Float) = setCategoryVolume(SoundCategory.AMBIENT, level)
     }
 
-    // TODO(CONVERT):
     // TODO(BREAKING): Make enums for most of these methods
     val video = object {
         //#if MC<=11202
         // TODO(BREAKING): Change return to int (and below methods argument type)
```
mattco98: Looks like this comment can be removed now

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Settings.kt
```diff
@@ -335,19 +339,28 @@ class Settings {
     }
 
     val chat = object {
+        // TODO(BREAKING) Switched to return a string for consistency with the setter
```
mattco98: It seems like the setter should just exclusively use the enum, similar to all of the methods above.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Renderer.kt
```diff
@@ -973,21 +1005,26 @@ object Renderer {
...
-    enum class DrawMode {
-        Lines,
-        LineStrip,
-        DebugLines,
-        DebugLineStrip,
-        Triangles,
-        TriangleStrip,
-        TriangleFan,
-        Quads;
+    enum class DrawMode(private val glMode: Int) {
```
mattco98, line 1016: Add breaking comment for removal of `Debug` members

mattco98 (changes requested):

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Settings.kt
mattco98: (To be clear, the comment on line 171. It isn't so obvious from the github preview)

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Player.kt
camnwalter: No, the naming changed so that IInventory switched to the Container interface in newer versions, and Container switched to AbstractContainerScreen.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/CTRenderPlayer.kt
camnwalter: Yeah I can do that.  There are a ton of layers other than those, should those be added as well?

mattco98: Hmm, maybe we should accept an object or something in that case

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/entity/Entity.kt
camnwalter: On the title screen the Entity would be null so this wouldn't  be an issue.

src/main/kotlin/com/chattriggers/ctjs/commands/Command.kt
camnwalter: I suggested they make a dsl option for commands a couple of months back, but there wasn't any reaction to it.  I could possibly try to do hacky stuff by adding annotations at runtime somehow but even then, I don't think it would work.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/CTRenderPlayer.kt
camnwalter: Also, should we have all of those layers as flags or just the ones that are useful to disable?

[Here's the list](https://nekoyue.github.io/ForgeJavaDocs-NG/javadoc/1.17.1/net/minecraft/client/renderer/entity/layers/package-summary.html) if you wanted to see them all, and it probably will keep increasing with more and more Minecraft versions.

src/main/kotlin/com/chattriggers/ctjs/commands/Command.kt
mattco98: Well it is what it is, we can see if it's possible later.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/CTRenderPlayer.kt
mattco98: Yeah that's a bit much. Seems like something we can add later. Maybe a TODO there would be nice.

PR #340, opened 2022-07-16
Update ForgeGradle and other small changes

camnwalter:


src/main/kotlin/com/chattriggers/ctjs/commands/Command.kt
```diff
@@ -68,7 +72,14 @@ class Command(
 
     fun register() {
         if (name in ClientCommandHandler.instance.commandMap.keys && !overrideExisting) {
-            throw IllegalArgumentException("Command with name $name already exists!")
+            ModuleManager.loaders.forEach {
```
mattco98: Instead of doing this, how about we make `Trigger::loader` internal and use that (since from what I can tell there is literally no reason for `triggers` to be a list)

mattco98 (changes requested):

PR #341, opened 2022-07-19
Add image.destroy method

Soopyboo32:
At the moment images only get garbage collected after a ct load, this can cause a memory leak if the images are no longer used from the js side.

So the idea is before u remove references to an images in js side u call image.destroy() and that will clear all references to it in the CT code so that its able to be garbage collected.
It also deletes the gl texture that way its removed from GPU ram

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Image.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Image.kt
@@ -49,6 +49,15 @@ class Image constructor(var image: BufferedImage?) {
         return texture
     }
 
+    /**
+     * Clears the image from GPU memory and removes its references CT side
+     * that way it can be garbage collected if not referenced in js code.
+     */
+    fun destroy() {
+        texture.deleteGlTexture()
+        CTJS.images.remove(this)
+    }
+
     @SubscribeEvent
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Image.kt
```diff
@@ -49,6 +49,15 @@ class Image constructor(var image: BufferedImage?) {
         return texture
     }
 
+    /**
+     * Clears the image from GPU memory and removes its references CT side
+     * that way it can be garbage collected if not referenced in js code.
+     */
+    fun destroy(){
```
mattco98: Add a space before the brace

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Image.kt
```diff
@@ -49,6 +49,15 @@ class Image constructor(var image: BufferedImage?) {
...
+        texture.deleteGlTexture()
+        CTJS.images.remove(this)
```
mattco98, line 58: Since we are now iterating this list, it would be best to make `images` a set, since we don't need ordering.

mattco98 (changes requested):
Please squash your commits and format them similarly to other commits in this repo.

PR #342, opened 2022-07-22
Renderer: Convert drawStringInWorld to 1.17

camnwalter:
Also moved RenderWorld trigger to where it injects for Fabric.  Before, It was injecting slightly too late.

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Renderer.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Renderer.kt
@@ -33,9 +33,10 @@ import net.minecraft.client.renderer.RenderHelper
 //$$ import com.mojang.blaze3d.platform.Lighting
 //$$ import com.mojang.blaze3d.vertex.PoseStack
 //$$ import com.mojang.blaze3d.systems.RenderSystem
-//$$ import com.mojang.math.Quaternion
-//$$ import net.minecraft.client.CameraType
+//$$ import com.mojang.math.Matrix4f
+//$$ import com.mojang.math.Transformation
 //$$ import net.minecraft.client.gui.GuiComponent
+//$$ import net.minecraft.client.renderer.MultiBufferSource
 //$$ import net.minecraft.client.renderer.entity.EntityRendererProvider
 //#endif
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Renderer.kt
```diff
@@ -760,11 +775,11 @@ object Renderer {
      * @param renderBlackBox render a pretty black border behind the text
      * @param scale the scale of the text
      * @param increase whether to scale the text up as the player moves away
+     * @param visibleThroughWalls whether to show the text through walls
      */
     // TODO(BREAKING): Rename from drawString (in Tessellator)
     // TODO(BREAKING): Remove color argument. Users should use colorize()
-    // TODO(CONVERT): Currently does not work in 1.17
-    // TODO(CONVERT): Is the scale argument necessary? Does scale() affect this function?
+    // TODO(BREAKING): Add visibleThroughWalls argument. Not sure if this is breaking or not if you leave it out.
```
mattco98: This shouldn't be breaking, right? since it has a default

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Renderer.kt
```diff
@@ -775,47 +790,34 @@ object Renderer {
...
         val renderPos = getRenderPos(x, y, z)
 
         if (increase) {
-            //#if MC<=11202
             val distance = sqrt(renderPos.x * renderPos.x + renderPos.y * renderPos.y + renderPos.z * renderPos.z)
-            //#else
-            //$$ val distance = sqrt(renderPos.x() * renderPos.x() + renderPos.y() * renderPos.y() + renderPos.z() * renderPos.z())
-            //#endif
-            val multiplier = distance / 120f //mobs only render ~120 blocks away
+            val multiplier = distance / 6f
```
mattco98, line 800: What does changing this from `120` do

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Renderer.kt
camnwalter: Checking this right now.

Yeah never mind, it doesn't break.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Renderer.kt
camnwalter: Oh, well looking back, I actually missed something.  But these changes will make the scaling much more consistent between having `increase` true or false, and a scale of 1 is going to be much closer in size to a nametag than it was before.

The text in the pictures is a couple blocks to the left and up from that platform

Before: increase false
image

Before: increase true
image

After: increase false
image

After: increase true
image

PR #344, opened 2022-07-26
Convert final triggers to 1.17

camnwalter:
Converts and/or fixes typos adding:
- Tooltip
- GuiRender
- GuiKey
- GuiMouseClick
- GuiMouseRelease
- GuiMouseDrag
- ChatComponentClicked
- ChatComponentHovered
- GuiDrawBackground
- RenderTitle
- PostGuiRender (typo)
- PostRenderEntity (typo)
- NoteBlockChange (typo)


src/main/java/com/chattriggers/ctjs/mixins/gui/GuiIngameMixin.java
```diff
@@ -204,5 +220,33 @@ private void chattriggers_renderScoreboardTrigger(
     //$$         ci.cancel();
     //$$     }
     //$$ }
+    //$$
+    //$$ // @WrapWithCondition(
```
camnwalter, line 224: FYI: These methods are commented out since the current version of MixinExtras bundled with Essential only supports WrapWithCondition on void methods.  This hopefully can be uncommented out after the next Essential update.

PR #346, opened 2022-08-24
onGuiOpened register for Guis

DalwynWasTaken:
Allows you to create a [onGuiOpened] register on a specific gui, this also allows the passing of the gui as an argument

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/gui/Gui.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/gui/Gui.kt
@@ -18,7 +18,7 @@ abstract class Gui : GuiScreen() {
     private var onMouseReleased: RegularTrigger? = null
     private var onMouseDragged: RegularTrigger? = null
     private var onActionPerformed: RegularTrigger? = null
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/gui/Gui.kt
```diff
@@ -156,6 +158,19 @@ abstract class Gui : GuiScreen() {
...
+     * Arguments passed through to method:
+     * - the gui
+     *
+     * @param method the method to run
+     * @return the trigger
+     */
+    fun registerGuiOpened(method: Any) = run {
+        onGuiOpened = RegularTrigger(method, TriggerType.Other, getLoader())
+        onGuiOpened
+    }
```
mattco98: Let's remove the `Gui` from all of these identifiers, we're already in the Gui class so it's a bit redundant

mattco98 (changes requested):

mattco98:
Sorry to make you push twice, should have thought about this earlier, but if we're adding an opened trigger, we should also probably add a closed trigger.

PR #350, opened 2022-08-31
Added Player.isMoving() and a small bug fix

realfork:
Convenience method for checking if a player is moving

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Player.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Player.kt
@@ -196,6 +196,9 @@ object Player {
     fun getLightLevel(): Int = World.getWorld()?.getLight(getPlayer()?.position) ?: 0
 
     @JvmStatic
+    fun isMoving(): Boolean = getPlayer()?.let { it.movementInput.moveForward != 0F || it.movementInput.moveStrafe != 0F } ?: false
+
+    @JvmStatic
     fun isSneaking(): Boolean = getPlayer()?.isSneaking ?: false
 
     @JvmStatic
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Player.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Player.kt
@@ -196,7 +196,7 @@ object Player {
     fun getLightLevel(): Int = World.getWorld()?.getLight(getPlayer()?.position) ?: 0
 
     @JvmStatic
-    fun isMoving(): Boolean = getPlayer()?.let { it.movementInput.moveForward != 0F || it.movementInput.moveStrafe != 0F } ?: false
+    fun isMoving(): Boolean = getPlayer()?.movementInput?.let { it.moveForward != 0F || it.moveStrafe != 0F } ?: false
 
     @JvmStatic
     fun isSneaking(): Boolean = getPlayer()?.isSneaking ?: false
--- a/src/main/kotlin/com/chattriggers/ctjs/launch/plugin/guiIngameForge.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/launch/plugin/guiIngameForge.kt
@@ -39,7 +39,7 @@ fun injectRenderTitle() = inject {
         val displayedSubTitle = shadowField<String>()
 
         code {
-            if (displayedTitle.isNotEmpty() && displayedSubTitle.isNotEmpty()) {
+            if (displayedTitle.isNotEmpty() || displayedSubTitle.isNotEmpty()) {
                 val event = CancellableEvent()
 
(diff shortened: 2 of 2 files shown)


realfork:
**Changes:**

- RenderEntityModel trigger
- EntityCull trigger (Unsure about it, can be deleted)
- MotionUpdatePre + MotionUpdatePost (Not 100% sure what to name these since it isn't technically motion (onUpdateWalkingPlayer))
- Fix to RenderTitle trigger

mattco98:
Can you provide use cases for these? For example, if I understand it correctly, the movement trigger can easily be achieved with a tick trigger and an if statement, and I'm not sure what you would use the pre/post versions for, since they are not cancellable. And it seems like instead of adding triggers for model parts, it would be much more useful to add accessors for them to the various entity classes that could be used in the `renderEntity` trigger that already exists.

realfork:
Most of them are for convenience.

RenderEntityModel - ModelBase is painful to get even with accessors so it's to massively simplify that process.
Movement - Is nicer than writing a tick trigger with the function but it can be removed.
MotionUpdate triggers - For running code before and after movement packets are sent.

mattco98:
> RenderEntityModel - ModelBase is painful to get even with accessors so it's to massively simplify that process.

Right, which is why CT can potentially take that burden off of the users with accessors, even if they are "painful". If it is feasible, we should go that route instead. And the bar for feasible for CT code (.kt) is much much lower than it is for user modules. 

> Movement - Is nicer than writing a tick trigger with the function but it can be removed.

I agree with the removal, doesn't seem to have much motivation

> MotionUpdate triggers - For running code before and after movement packets are sent.

That's not a use case, that is what they do, which I can infer from the code. What do you plan on doing with them that cannot be done another way?

realfork:
Ok, I finally found the way to get main model so I added it to the entity wrapper and removed the trigger. I also removed the movement triggers.

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/entity/Entity.kt
```diff
@@ -151,6 +154,13 @@ open class Entity(val entity: MCEntity) {
      */
     fun getClassName(): String = entity.javaClass.simpleName
 
+    /**
+     * Gets the ModelBase of the entity
+     *
+     * @return the entity's main model
+     */
+    fun getMainModel(): ModelBase = (Client.getMinecraft().renderManager.getEntityRenderObject<net.minecraft.entity.Entity>(entity) as RendererLivingEntity).getMainModel()
```
mattco98: Can this cast fail?

mattco98: We probably want a simple wrapper around `ModelBase` which forwards methods to an underlying implementation so the user doesn't need to use obfuscated methods. See [`Particle`](https://github.com/ChatTriggers/ChatTriggers/blob/master/src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/entity/Particle.kt) for an example. Idk what `ModelBase` looks like but as in `Particle`, feel free to change method names or add helper methods if you want.

mattco98: We should have an `MCEntity` alias in [`TypeAliases.kt`](https://github.com/ChatTriggers/ChatTriggers/blob/master/src/main/kotlin/com/chattriggers/ctjs/utils/kotlin/TypeAliases.kt) so you don't have to write out the full package name

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/Player.kt
```diff
@@ -195,6 +195,9 @@ object Player {
     @JvmStatic
     fun getLightLevel(): Int = World.getWorld()?.getLight(getPlayer()?.position) ?: 0
 
+    @JvmStatic
+    fun isMoving(): Boolean = getPlayer()?.let { it.movementInput.moveForward != 0F || it.movementInput.moveStrafe != 0F } ?: false
```
mattco98: Small nitpick to shorten the line a bit:
```suggestion
    fun isMoving(): Boolean = getPlayer()?.movementInput?.let { it.moveForward != 0F || it.moveStrafe != 0F } ?: false
```

mattco98 (changes requested):
Looks much better, thanks for refactoring!

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/entity/Entity.kt
realfork: I don't think so?

realfork:
Sorry for abandoning this for a while, I removed the model base method since it doesn't actually get the model correctly. Should be ready now.

PR #351, opened 2022-09-02
Modules: Make required modules not deletable until its dependents are…

camnwalter:
… deleted

Also shows the modules which require it.

Example picture:
image

I just realized this doesn't wrap the text if there are too many modules that require it, but I can add that if you'd like.

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/engine/module/ModuleUpdater.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/engine/module/ModuleUpdater.kt
@@ -15,7 +15,6 @@ import net.minecraftforge.event.world.WorldEvent
 import net.minecraftforge.fml.common.eventhandler.SubscribeEvent
 import org.apache.commons.io.FileUtils
 import java.io.File
-import java.net.URL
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/chattriggers/ctjs/engine/module/ModuleUpdater.kt
```diff
@@ -98,18 +97,36 @@ object ModuleUpdater {
         }
     }
 
-    fun importModule(moduleName: String): List<Module> {
-        if (cachedModules.any { it.name == moduleName }) return emptyList()
+    fun importModule(moduleName: String, requiredBy: String? = null): List<Module> {
+        if (cachedModules.any {
+                return@any if (it.name == moduleName) {
```
mattco98: Unnecessary `return@any`, also could we pull this into a variable? I'm not a fan of huge if statement conditions like this

mattco98 (changes requested):
Imo the text wrapping is not really an issue, I wouldn't worry about it. Probably more trouble than it's worth until we start using Elementa for this

PR #352, opened 2022-09-11
Added RayTraceResult wrapper and rayTrace methods

ghost:


mattco98:
Two issues from looking at the implementation of `Entity::rayTrace`:

1. It looks like it only returns blocks. If this is the case, then the method should be renamed to reflect that (and also potentially add a method for ray tracing the entity if that doesn't already exist)
2. Is the distance parameter a max distance? That's what it looks like, but it's hard to tell. If it is, then it should be an optional argument and default to a large number (500 or something).

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/entity/Entity.kt
```diff
@@ -296,7 +300,57 @@ open class Entity(val entity: MCEntity) {
...
+            lookDirection.zCoord * maxDistance
+        )
+
+        val blockHit = rayTraceBlock(maxDistance, partialTicks)
+        var entityHit: MCEntity? = null
+        var hitVec: Vec3 = eyePosition
+
+        getWorld().getEntitiesWithinAABBExcludingEntity(
+            entity,
+            entity.entityBoundingBox.addCoord(
```
mattco98, line 330: Can we create the bounding box with `blockHit` if it exists? That way if we're looking at a block 10 blocks away, we scan a much smaller box

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/entity/Entity.kt
```diff
@@ -296,7 +300,57 @@ open class Entity(val entity: MCEntity) {
...
+                lookDirection.xCoord * maxDistance,
+                lookDirection.yCoord * maxDistance,
+                lookDirection.zCoord * maxDistance
+            )
+        ).forEach { entityFound ->
+            entityFound.entityBoundingBox.calculateIntercept(eyePosition, lookingAt)?.also { result ->
```
mattco98, line 336: Maybe I'm missing it, but is this guaranteed to return the closest entity? Like if you're looking through multiple, it seems like it'll just pick the one that happened to be returned last from `getEntitiesWithinAABBExcludingEntity`

src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/entity/Entity.kt
```diff
@@ -296,7 +300,57 @@ open class Entity(val entity: MCEntity) {
...
+                null,
+                MCBlockPos(hitVec)
+            ))
+    }
+
     override fun toString(): String {
         return "Entity{name=${getName()}, x=${getX()}, y=${getY()}, z=${getZ()}}"
     }
-}
+}
```
mattco98, line 356: Revert

mattco98 (changes requested):

PR #356, opened 2022-10-20
Fix NPE with RenderManager in drawString

Debuggingss:
I was not able to reproduce the issue as it never happened to me before. According to @Squagward this change fixes the issue. I have no idea why it's happening though.


DJtheRedstoner (changes requested):
This doesn't fix the underlying issue. Since renderManager is initialized once, if it ends up null the internal state is invalid. Maybe make it `lazy`?

DJtheRedstoner (approved):

PR #357, opened 2022-11-02
Add Font class for custom font rendering

realfork:


src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+
+        GL11.glBegin(GL11.GL_LINES)
+        GL11.glVertex2f(start.x, start.y)
+        GL11.glVertex2f(end.x, end.y)
+        GL11.glEnd()
+
+        GlStateManager.enableTexture2D()
+    }
+
+    private fun generateColors() {
```
mattco98, line 369: This method is complete gibberish. Please pick better variable names and add some comments to help the reader understand what is happening. Also, it looks like this (and `colorCodes`) can be static.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+        }
+    }
+
+    internal inner class CharacterData(var width: Float, var height: Float, private val textureId: Int) {
+        fun bind() = GlStateManager.bindTexture(textureId)
+    }
+
+    companion object {
+        private const val MARGIN = 4
+        private const val COLOR_INVOKER = '\u00a7'
```
mattco98, line 394: This seems like a strange name. Something like `COLOR_CODE_SYMBOL` would be more appropriate.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+                            if (shadow) 0.25F else 1F,
+                            if (shadow) 0.25F else 1F,
+                            if (shadow) 0.25F else 1F,
+                            a
+                        )
+                    }
+                }
+            } else {
+                if (character.code > 255) continue
+                if (obfuscated) character = (character.code + RANDOM_OFFSET).toChar()
```
mattco98, line 233: I'm not sure about this random offset. Presumably we would always want to pick a printable character. Best to check the MC source to see what it does.

mattco98:
Before I post a review, is there a reason we can't use Elementa's font utilities? Seems like both `FontRenderer` and `BasicFontRenderer` (not sure what the difference is) accepts a `java.awt.Font`.

realfork:
Maybe, but I seem to remember it's very strange to use and the fonts aren't as smooth as this. Will have a look though and double check in case it has been improved since I last tried it.

realfork:
I looked into this, the Elementa font renderer uses https://github.com/Chlumsky/msdfgen for generating distance fields to use instead of a .ttf meaning it wouldn't really be suitable for something like CT.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+            }
+
+            graphics.setRenderingHint(RenderingHints.KEY_TEXT_ANTIALIASING, RenderingHints.VALUE_TEXT_ANTIALIAS_ON)
+            graphics.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON)
+            graphics.setRenderingHint(RenderingHints.KEY_RENDERING, RenderingHints.VALUE_RENDER_QUALITY)
+            graphics.setRenderingHint(RenderingHints.KEY_FRACTIONALMETRICS, RenderingHints.VALUE_FRACTIONALMETRICS_ON)
+
+            graphics.drawString(character.toString() + "", MARGIN, fontMetrics.ascent)
+
+            val textureId = GlStateManager.generateTexture()
```
mattco98, line 78: It doesn't look like this has a corresponding `deleteTexture` call, which will leak a lot of memory. We should more or less do the same thing that `Image` does.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+        val pixels = IntArray(image.width * image.height)
+
+        image.getRGB(0, 0, image.width, image.height, pixels, 0, image.width)
+
+        val buffer = BufferUtils.createByteBuffer(image.width * image.height * 4)
+        for (y in 0 until image.height) {
+            for (x in 0 until image.width) {
+                val pixel = pixels[y * image.width + x]
+
+                buffer.put((pixel shr 16 and 0xFF).toByte())
```
mattco98, line 101: Please add clarifying parens when chaining bitwise operators (entire file)

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+                characterImage.height.toFloat(),
+                textureId
+            )
+        }
+
+        return characterData
+    }
+
+    private fun createTexture(textureId: Int, image: BufferedImage) {
+        val pixels = IntArray(image.width * image.height)
```
mattco98, line 92: Maybe I'm misunderstanding, but I don't think this intermediate array is necessary, since `BufferedImage` supports individual pixel access

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+                if (strikethrough) drawLine(
+                    Vector2f(0F, charData!!.height / 2f), Vector2f(
+                        charData.width, charData.height / 2f
+                    )
+                )
+                if (underlined) drawLine(
+                    Vector2f(0F, charData!!.height - 15), Vector2f(
+                        charData.width, charData.height - 15
+                    )
+                )
```
mattco98, line 248:
```suggestion
                if (strikethrough) {
                    drawLine(
                        Vector2f(0F, charData!!.height / 2f), 
                        Vector2f(charData.width, charData.height / 2f),
                    )
                }

                if (underlined) {
                    drawLine(
                        Vector2f(0F, charData!!.height - 15), 
                        Vector2f(charData.width, charData.height - 15),
                    )
                }
```

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+        if (!generatedCharData) return 0F
+
+        var width = 0f
+
+        var characterData = regularData
+        val length = text.length
+
+        for (i in 0 until length) {
+            val character = text[i]
+            val previous = if (i > 0) text[i - 1] else '.'
```
mattco98, line 275:
```suggestion
            val previous = text.getOrElse(i - 1, '.')
```

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+
+            if (previous == COLOR_INVOKER) continue
+            if (character == COLOR_INVOKER) {
+                when ("0123456789abcdefklmnor".indexOf(text.lowercase()[i + 1])) {
```
mattco98, line 279: Why not just compare against the character directly instead of changing it to an offset?

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
```
mattco98, line 316: Same here

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+            regularData = setup(arrayOfNulls(256), Font.PLAIN)
+            boldData = setup(arrayOfNulls(256), Font.BOLD)
+            italicsData = setup(arrayOfNulls(256), Font.ITALIC)
+
+            generatedCharData = true
+        }
+    }
+
+    private fun setup(characterData: Array<CharacterData?>, type: Int): Array<CharacterData?> {
+        generateColors()
```
mattco98, line 42: This doesn't seem like it needs to be called three times for every construction of `Font`

e: The `generateColors` call, to be clear. It's not super clear from the GitHub preview.

mattco98 (changes requested):
Did you derive this from somewhere else?

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+
+    private lateinit var regularData: Array<CharacterData?>
+    private lateinit var boldData: Array<CharacterData?>
+    private lateinit var italicsData: Array<CharacterData?>
+    private var generatedCharData = false
+
+    private val colorCodes = IntArray(32)
+
+    init {
+        val inputStream = FileInputStream("${Config.modulesFolder}/$path")
```
DJtheRedstoner, line 29: unclosed stream

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+
+            graphics.setRenderingHint(RenderingHints.KEY_TEXT_ANTIALIASING, RenderingHints.VALUE_TEXT_ANTIALIAS_ON)
+            graphics.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON)
+            graphics.setRenderingHint(RenderingHints.KEY_RENDERING, RenderingHints.VALUE_RENDER_QUALITY)
+            graphics.setRenderingHint(RenderingHints.KEY_FRACTIONALMETRICS, RenderingHints.VALUE_FRACTIONALMETRICS_ON)
+
+            graphics.drawString(character.toString() + "", MARGIN, fontMetrics.ascent)
+
+            val textureId = GlStateManager.generateTexture()
+            createTexture(textureId, characterImage)
```
DJtheRedstoner, line 79: Creating a texture for each glyph is not very optimized. Consider creating a glyph atlas.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+        for (i in 0 until length) {
+            var character = text[i]
+            val previous = if (i > 0) text[i - 1] else '.'
+
+            if (previous == COLOR_INVOKER) continue
+            if (character == COLOR_INVOKER) {
+                var index = "0123456789abcdefklmnor".indexOf(text.lowercase()[i + 1])
+
+                when {
+                    index < 16 -> {
```
DJtheRedstoner, line 194: A lot of magic numbers here, can this be done in a nicer way?

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+                buffer.put((pixel shr 24 and 0xFF).toByte())
+            }
+        }
+        (buffer as Buffer).flip()
+
+        GlStateManager.bindTexture(textureId)
+        GL11.glTexParameteri(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_MIN_FILTER, GL11.GL_NEAREST)
+        GL11.glTexParameteri(GL11.GL_TEXTURE_2D, GL11.GL_TEXTURE_MAG_FILTER, GL11.GL_NEAREST)
+
+        GL11.glTexImage2D(
```
DJtheRedstoner, line 113: The majority of this function can be eliminated. By using an internal format of `GL_RGBA8` (`GL_RBGA` would work too, but sized formats are preferred these days), a format of `GL_BGRA` and a type of `GL_UNSIGNED_INT_8_8_8_8_REV`, you can directly pass the `pixels` array to opengl, without any need for transformation.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+    private var generatedCharData = false
+
+    private val colorCodes = IntArray(32)
+
+    init {
+        val inputStream = FileInputStream("${Config.modulesFolder}/$path")
+        this.font = Font.createFont(0, inputStream).deriveFont(Font.PLAIN, size)
+
+        Client.scheduleTask {
+            regularData = setup(arrayOfNulls(256), Font.PLAIN)
```
DJtheRedstoner, line 33: This seems a bit weird, couldn't the array just be constructed inside `setup`? Should also allow the elements to be non-nullable types.

src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Font.kt
```diff
@@ -0,0 +1,397 @@
...
+                red /= 4
+                green /= 4
+                blue /= 4
+            }
+
+            colorCodes[i] = red and 255 shl 16 or (green and 255 shl 8) or (blue and 255)
+        }
+    }
+
+    internal inner class CharacterData(var width: Float, var height: Float, private val textureId: Int) {
```
DJtheRedstoner, line 388: can this be `private`? does it need to be `inner`?

DJtheRedstoner (changes requested):
A few general points:
- this doesn't support unicode characters, is this something we want to support?
- would this work in ports to modern versions? or would the restrictions on the awt api prevent this?
- a lot of this reads like a java class (was it converted by intellij? :P), i can give more detailed comments on this if needed

mattco98:
Feel free to reopen later if you want to work on this again

PR #359, opened 2022-11-12
TextComponent: Don't use Forge's default link detection

camnwalter:
Previously, all links would be converted to have blue text and underline.  This is ugly and unneeded.

This allows for links with and without protocols but doesn't allow IPs.

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/message/TextComponent.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/message/TextComponent.kt
@@ -2,7 +2,10 @@ package com.chattriggers.ctjs.minecraft.objects.message
 
 import com.chattriggers.ctjs.minecraft.libs.ChatLib
 import com.chattriggers.ctjs.utils.kotlin.*
-import net.minecraftforge.common.ForgeHooks
+import net.minecraft.event.ClickEvent
+import net.minecraft.util.ChatStyle
+import net.minecraft.util.EnumChatFormatting
+import java.net.URI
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/message/TextComponent.kt
```diff
@@ -228,4 +231,75 @@ class TextComponent {
...
+                    string[i] == 'r' -> style = ChatStyle()
+                }
+            } else if (URL_REGEX.matchesAt(string, i)) {
+                // add the previous component as a sibling of the base component
+                val prevText = MCBaseTextComponent(buffer.toString()).apply { chatStyle = style.createDeepCopy() }
+                buffer.clear()
+                comp.appendSibling(prevText)
+
+                val link = URL_REGEX.matchAt(string, i)!!.value
+                i += link.length - 1
```
mattco98, line 266: Why the `- 1` here? Maybe I'm dumb but that doesn't seem necessary, if the link is 10 characters long, we'd want to start searching 10 characters after the start.

camnwalter: I originally had it without the -1 but it was skipping the character right after the link.  Pretty sure it's due to the fact the index is incremented again at the end of the loop, but yeah this is what worked.

mattco98: Ahh right that makes sense

PR #360, opened 2022-11-20
Added index check to Player.setHeldItem

kerbybit:
I have to fix my intellij so I honestly have no idea if this works so I figure its better to open a pull request


mattco98:
This is technically breaking, and so unfortunately will need to wait until the next major version release

PR #363, opened 2023-01-16
ChatLib: Fix editChat overflowing & a bunch of other issues

camnwalter:
Now only modifies the chatLines field, which when calling  GuiNewChat::refreshChat, will update all drawn lines accordlingly.


mattco98:
Can we remove `drawnChatLines` from the access transformer?

camnwalter:
Updated!

PR #364, opened 2023-01-18
Triggers: Add EntityRenderTrigger with dedicated class filter.

Marvinschs:
This is based on the PacketTrigger, since often when the `renderEntity` Trigger is used there is just a check if the entity which is rendered has the right class (and returned otherwise) before any modification happens. This leads to a lot of calls that dont have any use except returning


mattco98:
Good idea, thanks!

PR #368, opened 2023-01-23
Renderer: Various changes around scaling and Strings

Marvinschs:
- Added different `Renderer.scale(scaleXY)` and `Renderer.scale(scaleX, scaleY)` methods. (both types support Float, Double and Int)
- Added `getStringHeight()`. (wich returns the current `Font_Height` respective to yScale)
- Changed `getStringWidth(text: String)` to now reflext scale as well.
- Added `resetScale()` method to reset the scale to default.
- Added methods to get both scales.
- Added some minor helper methods.

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Renderer.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/renderer/Renderer.kt
@@ -165,8 +165,7 @@ object Renderer {
     }
 
     @JvmStatic
-    @JvmOverloads
-    fun scale(scaleX: Float, scaleY: Float = scaleX) = apply {
+    fun scale(scaleX: Float, scaleY: Float) = apply {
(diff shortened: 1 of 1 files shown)


Marvinschs:
- Changed the `getStringWidth(text)` to `getStringWidth(text, includeScale = false)` to avoid breaking anything while providing an option to include scale if wanted
- Removed unnecessary methods to just use `Number` parameter type
- Changed `resetScale()` to now actually reset the scale not just for the `Renderer` object but also for the `GlStateManager` used for drawing in `Renderer`

Marvinschs:
- Added `Renderer.pushMatrix()`
- Added `Renderer.popMatrix()` wich resets the internal scale
- Changed the default value of both scales to be `null` but methods now have a default scale of 1f
- Changed `resetScale()` to reset both scales to `null`

Marvinschs:
will close this pr and either make a new one or just not care at all anymore

PR #369, opened 2023-01-25
-Add `getStringHeight(string)` method

Marvinschs:


mattco98:
Please squash your commits and remove the leading dash

Marvinschs:
I actually will close this since it has like no use

PR #370, opened 2023-02-02
TextComponent: Fix invalid links causing an error…

camnwalter:
also fixes styles not getting reset each new line

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/message/TextComponent.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/message/TextComponent.kt
@@ -6,6 +6,7 @@ import net.minecraft.event.ClickEvent
 import net.minecraft.util.ChatStyle
 import net.minecraft.util.EnumChatFormatting
 import java.net.URI
+import java.net.URISyntaxException
 
 class TextComponent {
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/message/TextComponent.kt
```diff
@@ -269,13 +270,17 @@ class TextComponent {
...
+                    try {
+                        chatStyle.chatClickEvent = ClickEvent(
+                            ClickEvent.Action.OPEN_URL, if (URI(link).scheme == null) {
+                                "http://$link"
+                            } else {
+                                link
+                            }
+                        )
+                    } catch (ignored: URISyntaxException) {
+                        // this will throw if there is bad uri syntax (e.g. a : with no port)
```
mattco98, line 282: Should we at least print this error in the console?

camnwalter: I don't really see the point since it most likely wasn't supposed to be a link in the first place.  Just seems like it might spam the console for not really important "errors."

Also, the module creator should notice it didn't work since it isn't clickable if it fails.

mattco98: Fair enough

PR #371, opened 2023-02-06
Triggers: Added ClassTrigger

Marvinschs:
Basically the old PR with a few changes:
- Added the old 
```js
setPacketClass(PacketA.class)
setPacketClasses([PacketA.class, PacketB.class])
```
methods but marked them as deprecated to not break modules due to the changes
- code cleanup to make it more readable


src/main/kotlin/com/chattriggers/ctjs/triggers/ClassTrigger.kt
```diff
@@ -0,0 +1,62 @@
...
+     */
+    fun setClasses(classes: List<Class<*>>) = apply { triggerClasses = classes }
+    override fun trigger(args: Array<out Any?>) {
+        val placeholder = evalTriggerType(args)
+        if(triggerClasses.isEmpty() || triggerClasses.any { it.isInstance(placeholder) })
+            callMethod(args)
+    }
+
+    private fun evalTriggerType(args: Array<out Any?>): Any {
+        return when(this.triggerType) {
```
mattco98: This is an _extremely_ hot function, so this `when` statement may be a performance concern. Let's make this class abstract and add an abstract method that subclasses can override to implement this portion of logic.

src/main/kotlin/com/chattriggers/ctjs/triggers/ClassTrigger.kt
```diff
@@ -0,0 +1,62 @@
+package com.chattriggers.ctjs.triggers
+
+import com.chattriggers.ctjs.engine.ILoader
+import com.chattriggers.ctjs.minecraft.wrappers.entity.Entity
+import com.chattriggers.ctjs.minecraft.wrappers.entity.TileEntity
+import com.chattriggers.ctjs.utils.kotlin.MCEntity
+import com.chattriggers.ctjs.utils.kotlin.MCTileEntity
+import net.minecraft.network.Packet
+
+class ClassTrigger(method: Any,private val triggerType: TriggerType, loader: ILoader): Trigger(method, triggerType, loader) {
```
mattco98: Missing a space before `private val triggerType` and before `: Trigger`

src/main/kotlin/com/chattriggers/ctjs/triggers/ClassTrigger.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    @Deprecated("Use the better named method", ReplaceWith("setClasses([A.class, B.class, C.class])"))
+    fun setPacketClasses(classes: List<Class<*>>) = setClasses(classes)
+
+    /**
+     * Alias for `setClasses([A.class, B.class])`
+     *
+     * @param clazz The class for which this trigger should run for
+     */
+    fun setClass(clazz: Class<*>?) = setClasses(listOfNotNull(clazz))
```
mattco98: Let's rename the class to `ClassFilterTrigger` and these methods to `setFilteredClass` and `setFilteredClasses`

src/main/kotlin/com/chattriggers/ctjs/triggers/ClassTrigger.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.chattriggers.ctjs.minecraft.wrappers.entity.Entity
+import com.chattriggers.ctjs.minecraft.wrappers.entity.TileEntity
+import com.chattriggers.ctjs.utils.kotlin.MCEntity
+import com.chattriggers.ctjs.utils.kotlin.MCTileEntity
+import net.minecraft.network.Packet
+
+class ClassTrigger(method: Any,private val triggerType: TriggerType, loader: ILoader): Trigger(method, triggerType, loader) {
+    private var triggerClasses: List<Class<*>> = emptyList()
+
+    @Deprecated("Use the better named method", ReplaceWith("setClass(MyClass.class)"))
```
mattco98:
```suggestion
    @Deprecated("Prefer setFilteredClass", ReplaceWith("setFilteredClass(MyClass.class)"))
```

src/main/kotlin/com/chattriggers/ctjs/triggers/ClassTrigger.kt
```diff
@@ -0,0 +1,62 @@
...
+    fun setPacketClass(clazz: Class<*>?) = setClasses(listOfNotNull(clazz))
+
+    @Deprecated("Use the better named method", ReplaceWith("setClasses([A.class, B.class, C.class])"))
```
mattco98:
```suggestion
    @Deprecated("Prefer setFilteredClasses", ReplaceWith("setFilteredClasses([A.class, B.class, C.class])"))
```

src/main/kotlin/com/chattriggers/ctjs/triggers/ClassTrigger.kt
```diff
@@ -0,0 +1,62 @@
...
+    fun setClass(clazz: Class<*>?) = setClasses(listOfNotNull(clazz))
+
+    /**
+     * Sets which classes this trigger should run for. If the list is empty, it runs
+     * for every class.
+     *
+     * @param classes The classes for which this trigger should run for
+     * @return This trigger object for chaining
+     */
+    fun setClasses(classes: List<Class<*>>) = apply { triggerClasses = classes }
```
mattco98: Newline after this method

mattco98 (commented):
By the way, you don't need to open a new PR if you want to make changes, you can just amend your commits and force push (google is your friend here if you don't know how to do this).

src/main/kotlin/com/chattriggers/ctjs/triggers/ClassTrigger.kt
Marvinschs: How would you want the implementation of the subclasses done?
I think it would be either to create an extra package in the `triggers` package and put them in different files, which looks just clean and allows for even more custom modification for each of the filtered trigger classes or just have them all in one big file.
But im open to even better alternatives.

mattco98: None of that should be necessary, they would just extend this class instead of `Trigger` and override the abstract method.

Marvinschs: Yes I know.
What i meant was that all of these subclasses would either be inside the `ClassFilterTrigger` file or seperate.

mattco98: Right, and the answer is that nothing would change, i.e., the classes would stay where they are and not move inside this class.

Marvinschs:
updated to the changes requested and i also tested the functionality.

PR #372, opened 2023-02-19
NBT: Add fromJson methods

camnwalter:
This allows users to create NBT wrappers similar to how they can create objects in the other direction. Users can pass numbers with the letter of the type that they should be, "1s" will be parsed as the short value 1, etc.  This also attempts to pack arrays into NBTTag(Byte, Int)Array if every element in each is a valid byte/int (or a string parsed as one).

Marking this as draft for now due to some issues.

- Right now NBTTagList::fromJson will throw a class cast exception if it tries to make a list with only floats/bytes, where it will interpret and parse them as a NBTTag(Byte, Int)Array.  Right now, it seems that there should be a way for the user to specify if they want the typed tag array instead of the normal taglist, but I haven't found a good answer to that yet.
- I'm not entirely sure if NBT can have null values, so if not, then what do we do?  Throw an error or just ignore that key?  I'd say throwing an error would be better.
- Also as I'm writing this, I found that TagLists are only supposed to have 1 type inside them [from this link](http://web.archive.org/web/20110723210920/http://www.minecraft.net/docs/NBT.txt), so I'll have to add a check to that as well.
  - This is likely considered a breaking change, but we can consider adding the checks like MC does for setting values in the list would fix the bug that allows different types into the list

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/inventory/nbt/NBT.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/wrappers/inventory/nbt/NBT.kt
@@ -1,0 +1,125 @@
+package com.chattriggers.ctjs.minecraft.wrappers.inventory.nbt
+
+import com.chattriggers.ctjs.utils.kotlin.MCNBTBase
+import com.chattriggers.ctjs.utils.kotlin.MCNBTTagCompound
+import com.chattriggers.ctjs.utils.kotlin.MCNBTTagList
+import com.chattriggers.ctjs.utils.kotlin.getOption
+import net.minecraft.nbt.*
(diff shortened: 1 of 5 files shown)


mattco98:
Surely `fromObject` would be a better name right? Since `fromJson` implies you'd have to `JSON.stringify` first. Plus it fits nicely with `toObject`

camnwalter:
Yeah I was thinking that. For the NBTTagList, it would be fromArray then right?

mattco98:
Yeah that'd make the most sense

camnwalter:
Any thoughts on the first bullet point?

camnwalter:
Just updated from the results of the discussion.  Added a 2nd commit for the 3rd bullet point, however I can remove it if it is breaking.

PR #373, opened 2023-02-21
Triggers: Allow chat trigger to detect client-side messages

camnwalter:


camnwalter:
I can attempt to separate these into 2 triggers, if that makes more sense.  So, users can guarantee that something came from client or server, rather than it being ambiguous.

Edit: As of right now, I can not think of any way to completely separate them, as there is the issue where if you cancel a client-side trigger, it ends up canceling every chat message, not just client-side ones. The code for that is in <https://github.com/camnwalter/ChatTriggers/commit/d715c5b5509a7c07e8d06954641afd4678ae8b1c>

PR #374, opened 2023-03-15
KeyBind: fix registered triggers not unregistering correctly

Marvinschs:
This fixes the issue mentioned here: https://discord.com/channels/119493402902528000/688773480954855537/1073933069725224980
and the resulting conversation related to the same issue: https://discord.com/channels/119493402902528000/688773480954855537/1074072830024622100
The changes now return the Keybind itself from the functions since the `RegularTrigger` does not offer important ways to change the behaviour of the trigger except priority which I have never seen being used, especially in this case.
Also `Keybind.registerKeyPress(whateverFunc).registerKeyDown(anotherFunc)` would be possible now which was not the case before.


Marvinschs:
changed back to only adding the methods for unregistering.

PR #375, opened 2023-03-16
BlockPos: Add getAllInBox method

Marvinschs:
This provides an alternative to the method of using nested for loops for x, y, z and the calls to `World.getBlockAt(x, y, z)` just to get Blocks in a certain area.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ apply plugin: 'net.minecraftforge.gradle.forge'
 apply plugin: 'kotlin'
 apply plugin: 'org.jetbrains.dokka'
 
-version = "2.1.5-" + (
+version = "2.2.0-" + (
     mcVersion.toString().substring(0, 1) + "." +
         mcVersion.toString().substring(1, 3).replace("0", "") + "." +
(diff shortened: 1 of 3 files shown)


mattco98:
Why `BlockPos` instead of `Block`? Also this should probably be `World.getBlocksInCube` or something

Marvinschs:
`BlockPos` because it is the same in mc.

mattco98:
Well I guess I was referencing the PR description:

> This provides an alternative to the method of using nested for loops for x, y, z and the calls to World.getBlockAt(x, y, z) just to get Blocks in a certain area.

The above doesn't really seem true, since `getBlockAt` returns a `Block`, not a `BlockPos`

Marvinschs:
You are right `getBlockAt` returns a `Block` but before one had to use constructs like this:
```js
for (let x = 0; x < radius * 2; x++) {
      for (let z = 0; z < radius * 2; z++) {
        for (let y = 0; y < radius * 2; y++) {
          let pos = new BlockPos(cachedX - radius + x, cashedY - radius + y, cachedZ - radius + z)
          var block = World.getBlockAt(pos)
          //...
```

PR #376, opened 2023-03-18
DisplayLine: trigger changes

Marvinschs:
When registering a trigger for a DisplayLine the line is returned instead of the trigger wich allows for registering the trigger before adding the line to the Display.
Also added methods to unregister triggers form Display Lines, which also get unregistered from the loader.
Changed the type of reguired trigger for all triggers to `RegularTrigger` since all the methods only add this type of trigger

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/display/DisplayLine.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/display/DisplayLine.kt
@@ -6,7 +6,6 @@ import com.chattriggers.ctjs.minecraft.libs.renderer.Text
 import com.chattriggers.ctjs.minecraft.listeners.MouseListener
 import com.chattriggers.ctjs.minecraft.wrappers.Client
 import com.chattriggers.ctjs.triggers.RegularTrigger
-import com.chattriggers.ctjs.triggers.Trigger
 import com.chattriggers.ctjs.triggers.TriggerType
 import org.mozilla.javascript.NativeObject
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/display/DisplayLine.kt
```diff
@@ -116,24 +115,40 @@ abstract class DisplayLine {
         this.backgroundColor = backgroundColor
     }
 
-    fun registerClicked(method: Any) = run {
+    fun registerClicked(method: Any) = apply {
```
mattco98: Changing these `run`s to `apply`s is a breaking change

mattco98 (changes requested):

src/main/kotlin/com/chattriggers/ctjs/minecraft/objects/display/DisplayLine.kt
Marvinschs: That is unfortunate because i feel like it should be possible to register triggers for each line before they get added to the Display. Without this change the trigger methods need to be registered after the line was added to the Display which I think is counter intuitive and is a pain especially when trying to use the `setLines` function.

mattco98: I'm not sure how this change would affect that, but regardless we cannot introduce breaking changes no matter how nice they may be.

Marvinschs: Well this is basically how it had to be done before: 
```js
let line = new DisplayLine("This is a Test!")
let display = new Display().setRegisterType(DisplayHandler.RegisterType.RENDER_OVERLAY)

display.addLine(line)
line.registerClicked(whateverFunction())
```
This with multiple lines is just messy. (there are obviously ways to make it look good but you have to pay attention to where you add the line and where you add the trigger)
But now it would allow:
```js
let line = new DisplayLine("This is a Test!").registerClicked(whateverFunction())
let display = new Display().setRegisterType(DisplayHandler.RegisterType.RENDER_OVERLAY)

display.addLine(line)
```

The `addLine` function could be chained as well but this looks better.

PR #385, opened 2023-06-06
Color dependency modules in gray

ixnoahlive:
Added a key that colours modules gray in the module list if the isLibrary metadata key is set to true.

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/engine/module/Module.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/engine/module/Module.kt
@@ -27,7 +27,7 @@ class Module(val name: String, var metadata: ModuleMetadata, val folder: File) {
         )
         
         Renderer.drawStringWithShadow(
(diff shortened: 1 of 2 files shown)


mattco98:
Why would we not use the already present `isRequired` value?

ixnoahlive:
@mattco98 I was not informed this was for libraries, and thought it was for general dependencies and mods packaged together. I'll gladly edit this PR to fix the problem.

mattco98:
`isRequired` is true when the module is depended on by another, which is pretty much the definition of a library. 

Also, please do not close issues until they are actually resolved (i.e. until the PR is closed)

ixnoahlive:
Pull req. updated to fix this, sorry for all the trouble.

mattco98:
Please squash your commits. Also, do not submit code without testing it, or at the _very_ least compiling it. As it is, this code will not compile.

ixnoahlive:
Hm, was very certain this was functional. Looking into it. I am quite busy at the minute but I'll do my best to resolve these gripes.

PR #393, opened 2024-05-18
Add chat input utilities to ChatLib

ixnoahlive:
Adds the following utility functions to ChatLib for interacting with the Chat Input Box:
* getCaretPosition
  * Gets the position of the chat input caret
* setCaretPosition
  * Sets the position of the chat input caret
* getSelectedText
  * Gets the current selected text in the chat input
  
Useful for example in use for a symbol command that lets you paste symbols into your chat at a specific position, or more dynamic interactions with things like letting you select text and press a hotkey and have a mod do some funky stuff with that.

Generally just nice stuff that was previously behind reflection.

### DIFF
--- a/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/ChatLib.kt
+++ b/src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/ChatLib.kt
@@ -9,6 +9,7 @@ import com.chattriggers.ctjs.minecraft.wrappers.Player
 import com.chattriggers.ctjs.printToConsole
 import com.chattriggers.ctjs.utils.kotlin.times
 import net.minecraft.client.gui.ChatLine
+import net.minecraft.client.gui.GuiChat
 import net.minecraftforge.client.ClientCommandHandler
 import net.minecraftforge.client.event.ClientChatReceivedEvent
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/chattriggers/ctjs/minecraft/libs/ChatLib.kt
```diff
@@ -420,6 +421,41 @@ object ChatLib {
...
+     * Gets the position of the caret in the chat's input box.
+     * @return the zero-index caret position, -1 if not in chat
+     */
+    @JvmStatic
+    fun getCaretPosition(): Int {
+        if (Client.getMinecraft().currentScreen is GuiChat) {
+            return (Client.getMinecraft().currentScreen as GuiChat).inputField.cursorPosition
+        } else {
+            return -1
+        }
```
mattco98:
```suggestion
        return (Client.getMinecraft().currentScreen as? GuiChat)?.inputField?.cursorPosition ?: -1
```

Similar suggestions for the other methods as well (though I can't remember if this syntax works for assignment)

mattco98 (changes requested):

sinender (approved):

PR #396, opened 2024-06-05
Fix importing and other web requests on older Java versions

MisterCheezeCake:
This PR fixes web requests and module importing on versions of Java 8 which don't have modern root CAs shipped with them, causing module importing to fail with an unclear error message. I've added a new keystore `ctjskeystore.jks`, which was migrated from the root CAs in a modern Java 8 build. I ave also modified `CTJS.makeWebRequest` to use this keystore, which fixes all web requests the mod makes. 

In my testing, this fixes importing on the included Java version from the launcher and does not cause issues on modern Java versions either.


src/main/kotlin/com/chattriggers/ctjs/utils/SSLUtil.kt
```diff
@@ -0,0 +1,27 @@
+package com.chattriggers.ctjs.utils
+
+import java.security.KeyStore
+import javax.net.ssl.KeyManagerFactory
+import javax.net.ssl.SSLContext
+import javax.net.ssl.TrustManagerFactory
+object SSLUtil {
```
mattco98: This doesn't need it's own class imo. It should just be a lazy field on the CTJS class:

```kotlin
class CTJS {
    companion object {
        internal val sslContext by lazy {
            // ...
        }
    }
}
```

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
```diff
@@ -26,10 +27,12 @@ import net.minecraftforge.fml.common.event.FMLPreInitializationEvent
 import java.io.File
 import java.net.URL
 import java.net.URLConnection
+import javax.net.ssl.HttpsURLConnection
 import java.security.MessageDigest
 import java.util.*
 import kotlin.concurrent.thread
 
+
```
mattco98, line 39: Revert

src/main/kotlin/com/chattriggers/ctjs/utils/SSLUtil.kt
```diff
@@ -0,0 +1,27 @@
...
+            myKeyStore.load(SSLUtil::class.java.getResourceAsStream("/ctjskeystore.jks"), "changeit".toCharArray())
+            val kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm())
+            val tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm())
+            kmf.init(myKeyStore, null)
+            tmf.init(myKeyStore)
+            ctx = SSLContext.getInstance("TLS")
+            ctx?.init(kmf.keyManagers, tmf.trustManagers, null)
+        } catch (e: Exception) {
+            println("CTJS failed to load keystore. Web requests may fail on older Java versions")
+            e.printStackTrace()
```
mattco98: These should print to console, like most other errors in the mod

mattco98 (changes requested):
Can you share how you generated the key file?

MisterCheezeCake:
> Can you share how you generated the key file?

Thanks for the review suggestions. I'll make and test those changes later today.

As for how I generated the key file, I ran the following command (with the actual paths)

JDK8-U51/bin/keytool -importkeystore -srckeystore MODERN-TEMURIN-OPENJDK8/lib/security/cacerts -destkeystore ctjskeystore.jks

MisterCheezeCake:
I've made the suggested changes

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
```diff
@@ -111,4 +121,27 @@ object CTJS {
...
+        } catch (e: Exception) {
+            ModuleManager.generalConsole.println("Failed to load keystore. Web requests may fail on older Java versions", LogType.WARN)
+            ModuleManager.generalConsole.println(e.stackTrace, LogType.ERROR)
+            null
+        }
+
+    }
+    fun getSSLContext(): SSLContext? {
+        return sslContext;
+    }
```
mattco98: This isn't necessary, the field isn't used outside of the class

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
```diff
@@ -87,11 +93,15 @@ object CTJS {
 
         registerHooks()
     }
-
-    fun makeWebRequest(url: String): URLConnection = URL(url).openConnection().apply {
-        setRequestProperty("User-Agent", "Mozilla/5.0 (ChatTriggers)")
-        connectTimeout = 3000
-        readTimeout = 3000
+    fun makeWebRequest(url: String): URLConnection {
```
mattco98: Let's pay a bit more attention to the formatting here. Newlines before functions, 4-space indentation, etc.

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
```diff
@@ -111,4 +121,27 @@ object CTJS {
...
+        try {
+            val myKeyStore = KeyStore.getInstance("JKS")
+            myKeyStore.load(CTJS::class.java.getResourceAsStream("/ctjskeystore.jks"), "changeit".toCharArray())
+            val kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm())
+            val tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm())
+            kmf.init(myKeyStore, null)
+            tmf.init(myKeyStore)
+            var ctx = SSLContext.getInstance("TLS")
+            ctx?.init(kmf.keyManagers, tmf.trustManagers, null)
+            ModuleManager.generalConsole.println("SSL Context Initialized")
```
mattco98: This print is not necessary

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
```diff
@@ -111,4 +121,27 @@ object CTJS {
...
+            ctx
+        } catch (e: Exception) {
+            ModuleManager.generalConsole.println("Failed to load keystore. Web requests may fail on older Java versions", LogType.WARN)
+            ModuleManager.generalConsole.println(e.stackTrace, LogType.ERROR)
```
mattco98: Messages are printed to console as follows: 
```kotlin
// For strings
"Module $directory has invalid metadata.json".printToConsole(logType = LogType.whatever)

// For errors
e.printTraceToConsole()
```

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
```diff
@@ -87,11 +93,15 @@ object CTJS {
...
-
-    fun makeWebRequest(url: String): URLConnection = URL(url).openConnection().apply {
-        setRequestProperty("User-Agent", "Mozilla/5.0 (ChatTriggers)")
-        connectTimeout = 3000
-        readTimeout = 3000
+    fun makeWebRequest(url: String): URLConnection {
+            val connection = URL(url).openConnection()
+            connection.setRequestProperty("User-Agent", "Mozilla/5.0 (ChatTriggers)")
+            if (connection is HttpsURLConnection && sslContext != null) {
+                connection.sslSocketFactory = sslContext!!.socketFactory
```
mattco98: I don't think this null assertion should be necessary? `sslContext` is `val`, so it can't be reassigned

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
```diff
@@ -26,8 +27,13 @@ import net.minecraftforge.fml.common.event.FMLPreInitializationEvent
 import java.io.File
 import java.net.URL
 import java.net.URLConnection
+import java.security.KeyStore
+import javax.net.ssl.HttpsURLConnection
```
mattco98, line 31: These imports don't look sorted; is this how IntelliJ imported them?

mattco98 (changes requested):
Please squash your fixup commits into the original commit

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
MisterCheezeCake: I added this function so modules would be able to use the up to date SSL Context in their requests if they wanted, without being able to modify it or needing to manage it themselves. I can remove it if you don't think it should be there though.

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
MisterCheezeCake: It should theoretically not be necessary, but the compiler errors when it's not there

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
MisterCheezeCake: It is, is there a specific convention for sorting imports I should follow?

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
mattco98: Fair enough, must be something to do with property delegation then

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
mattco98: If IntelliJ doesn't change it then it's correct

src/main/kotlin/com/chattriggers/ctjs/CTJS.kt
mattco98: I think I'd rather keep this an implementation detail, the `makeWebRequest` function is already public so they can use that to get access to the functionality

mattco98: False brought up a good point that axios and requestv2 will probably need access to this. However, there's still no need to make it its own function, just make the field public. And you might as well add a `@JvmStatic` annotation while you're at it.

MisterCheezeCake:
Made requested changes and squashed commits

MisterCheezeCake:
As requested, sslContext is now at the top of the file since fields should go before methods

PR #400, opened 2024-07-12
StepTrigger: avoid getting system time for each thing

DocilElm:
something about making step trigger slightly better or something 
idk im not good at kt


src/main/kotlin/com/chattriggers/ctjs/triggers/StepTrigger.kt
```diff
@@ -37,15 +37,17 @@ class StepTrigger(method: Any, loader: ILoader) : Trigger(method, TriggerType.St
     }
 
     override fun trigger(args: Array<out Any?>) {
+        val currentTime = args.firstOrNull() as Long? ?: return
```
mattco98: This will always have a long in the first position, so no need to handle anything else:
```suggestion
        val currentTime = args[0] as Long
```

mattco98 (changes requested):

